PR #17, opened 2022-03-14
Updates for new processing methodology

travisdesell:
You can now run:

sh run_process_upload.sh [optional upload id]

If run with an upload_id, it will remove and re-import all data for that upload.  If not it will run in daemon mode as before.  The code will now also create emails for every upload processed.

### DIFF
--- a/db/create_tables.php
+++ b/db/create_tables.php
@@ -7,6 +7,8 @@ $cwd[__FILE__] = dirname($cwd[__FILE__]);
 require_once($cwd[__FILE__] . "/my_query.php");
 
 $drop_tables = false;
+$update_2022_02_17 = true;
+$update_turn_to_final = true;
 
 //need to drop and reload these tables for 2020_05_16 changes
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/ngafid/MavenStarter.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class MavenStarter {
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+
+        MavenCli cli = new MavenCli();
+        //cli.doMain(new String[]{"clean", "install"}, "project_dir", System.out, System.out);
+        cli.doMain(new String[]{"exec:java", "-Dexec.mainClass='org.ngafid.ProcessUpload'"}, "/ngafid/ngafid2.0", System.out, System.out);
+
+
```
drc3203, line 21: Is there a reason for the commented out code to stay? If so, we might want a comment saying why it would be uncommented

db/create_tables.php
```diff
@@ -7,6 +7,8 @@
 require_once($cwd[__FILE__] . "/my_query.php");
 
 $drop_tables = false;
```
drc3203, line 9: Just double checking, these booleans and their use in the conditionals are the only changes to this, right?

drc3203 (approved):
I think it looks pretty good, I left a few comments for clarification

aidanlabella:
Everything looks good to me. Would we want to possibly update the `update_tables.php` so that we can add compatibility in our local environments?

aidanlabella (approved):

PR #18, opened 2022-03-16
Updated sending email to send to admins as BCC

travisdesell:
Since we have a few people as admins that will get NGAFID emails, this update will send those emails out as BCC.


drc3203 (approved):

aidanlabella:
looks good!

aidanlabella (commented):
approving

PR #21, opened 2022-05-13
Event Description Tooltips

AarC10:


package.json
```diff
@@ -44,6 +44,7 @@
         "plotly.js": "^1.58.4",
         "popper": "^1.0.1",
         "react-bootstrap": "^1.6.1",
+        "react-tooltip": "^4.2.21",
```
aidanlabella, line 47: I believe I already have react-tooltips in here somewhere - see the saving filters feature on the flights page

src/main/javascript/trends.js
```diff
@@ -43,6 +44,8 @@ var eventCounts = {};
 var eventFleetPercents = {};
 var eventNGAFIDPercents = {};
 
+var tooltipText = "";
```
aidanlabella, line 47: Make this a part of the TrendsPage `state` or `props`

src/main/javascript/severities.js
```diff
@@ -307,6 +309,35 @@ class SeveritiesPage extends React.Component {
...
+                    text = response + "";
+                },
+                error : function(jqXHR, textStatus, errorThrown) {
+                    errorModal.show("Error Getting Event Description", errorThrown);
+                },
+                async: false
+            });
+
+        console.log("returning text: " + text);
+        return text;
```
aidanlabella, line 338: I would add a set/map in the `state` that contains the description. You can create a set that has keys as the event name that map to their respective event description. Then you can just check to see if it exists - if not we would query the server and populate that key. This would make the class cache the descriptions it already has asked the server for, cutting down on network resources.

src/main/javascript/trends.js
```diff
@@ -577,6 +580,36 @@ class TrendsPage extends React.Component {
...
+                url: '/protected/get_event_description',
+                data : {eventName : eventName},
+                dataType : 'json',
+                success : function(response) {
+                    console.log("received response: " + response);
+
+                    $('#loading').hide();
+
+                    if (response.err_msg) {
+                        errorModal.show(response.err_title, response.err_msg);
```
aidanlabella, line 598: Appreciate the spacing here - makes code easy to read :)

aidanlabella (changes requested):
Just needs a few minor changes - otherwise, well done!

src/main/javascript/severities.js
travisdesell: Yup this would be great, i don't think we want to be pinging the server everytime we mouseover a tooltip (especailly if we've already grabbed it).

src/main/javascript/trends.js
travisdesell: So it looks like get_event_description is something we're going to want to reuse across multiple webpages, so instead of having the function copied, we should have it in a single spot that we import from.

package.json
AarC10: I installed a package, not realizing that there was already another way of creating a tooltip.

travisdesell (approved):

PR #22, opened 2022-05-14
Tooltips for Flights Page and Updated Human Readable String

AarC10:


src/main/java/org/ngafid/events/EventDefinition.java
```diff
@@ -637,10 +637,17 @@ public static ArrayList<String> getAllNames(Connection connection, int fleetId)
      */
     public String toHumanReadable() {
         String text = "";
+
+        if (name.matches("^[AEIOU].*")) {
+            text = "An ";
+        } else {
+            text = "A ";
+        }
```
aidanlabella, line 645: Just a suggestion, you could do this in a ternary statement (i.e. `String text = (name.matches("^[AEIOU].*") ? "An " : "A ");`)

aidanlabella (approved):

src/main/java/org/ngafid/events/EventDefinition.java
AarC10: Nice catch. Thanks

PR #31, opened 2022-07-10
Fix Scan Eagle in Cesium

AarC10:


src/main/java/org/ngafid/routes/GetNgafidCesium.java
```diff
@@ -150,6 +150,9 @@ public Object handle(Request request, Response response) {
                 StringTimeSeries date = StringTimeSeries.getStringTimeSeries(connection, flightIdNewInteger, "Lcl Date");
                 StringTimeSeries time = StringTimeSeries.getStringTimeSeries(connection, flightIdNewInteger, "Lcl Time");
 
+                System.out.println(date.size());
```
aidanlabella, line 153: For consistency, can you change these to a Log.info()?

PR #32, opened 2022-07-15
Created new CSV writers for Aidan's research project

aidanlabella:
Also added the option for users to export CSVs using the generated option on the flights page.

### DIFF
--- a/src/main/java/org/ngafid/flights/CachedCSVWriter.java
+++ b/src/main/java/org/ngafid/flights/CachedCSVWriter.java
@@ -1,6 +1,8 @@
 package org.ngafid.flights;
 
 import java.io.File;
+import java.io.FileOutputStream;
+import java.io.FileWriter;
 import java.io.InputStream;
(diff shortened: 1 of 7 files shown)


src/main/java/org/ngafid/flights/CachedCSVWriter.java
```diff
@@ -0,0 +1,174 @@
...
+
+                fos.write(toBinaryData());
+                fos.close();
+            } catch (IOException ie) {
+                ie.printStackTrace();
+            }
+        } else {
+            // This should not happen!
+            return;
+        }
```
AarC10, line 162: Is there a specific reason to include the else block in both your Cached and Generated CSV Writers, if it just exits the function regardless?

AarC10 (approved):
Looks fine to me. Only thing is that I am curious about the use of the else in both CSV Writer implementations. Good job though.

travisdesell (approved):

PR #36, opened 2022-07-27
Merge just the code for spin events

aidanlabella:
This is just the code for the spin events. I will need to run the calculations on the VM at some point once it is up to date with `main`. The event definitions table should also be updated at some point.


AarC10 (approved):
Seems good

travisdesell (approved):

PR #39, opened 2022-08-10
Low Fuel Average Custom Event

AarC10:


travisdesell (changes requested):

src/main/java/org/ngafid/FindLowEndingFuelEvents.java
```diff
@@ -0,0 +1,168 @@
...
+
+            try {
+                List<Upload> uploads = Upload.getUploads(connection, fleetID);
+                uploadSize = uploads.size();
+
+                for (Upload upload : uploads) {
+                    findLowEndFuelEventsInUpload(upload);
+                }
+            } catch (SQLException e) {
+                e.printStackTrace();
```
travisdesell, line 157: Could we get an exit() here?

src/main/java/org/ngafid/common/TSNIterator.java
```diff
@@ -0,0 +1,24 @@
+package org.ngafid.common;
```
travisdesell: If we're not using the time series node code please remove.

travisdesell (changes requested):

src/main/java/org/ngafid/FindLowEndingFuelEvents.java
```diff
@@ -0,0 +1,173 @@
...
+public class FindLowEndingFuelEvents {
+    public static final Connection connection = Database.getConnection();
+    public static final Logger LOG = Logger.getLogger(FindLowEndingFuelEvents.class.getName());
+    private static Map<Integer, EventDefinition> eventDefs = new HashMap<>();
+    private static Map<Integer, Double> thresholds = new HashMap<>();
+
+
+    public static void findLowEndFuelEventsInUpload(Upload upload) {
+        try {
+            String whereClause = "upload_id = " + upload.getId() + " AND insert_completed = 1 AND NOT EXISTS (SELECT flight_id FROM events WHERE id = -4 OR id = -5 OR id = -6)";
```
travisdesell, line 29: look at the where statement on line 598 in CalculateProximity.java and re-do this.

travisdesell: also, make the -4, -5, -6 static variables (not hard coded).

travisdesell (changes requested):

travisdesell (approved):

PR #41, opened 2022-08-26
Aidan/spin events

aidanlabella:
Fixed issue where there is a duplicate key exception after restarting the `find_spin_events.sh` script after a crash.


src/main/java/org/ngafid/FindSpinEvents.java
```diff
@@ -47,9 +47,31 @@ public static void findSpinEventsInUpload(Upload upload) {
...
+
+        query.executeUpdate();
+
+        sql = "DELETE FROM flight_processed WHERE flight_id = ? AND event_definition_id IN (?,?)";
+        query = connection.prepareStatement(sql);
+
+        query.setInt(1, flight.getId());
+        query.setInt(2, LOW_ALTITUDE_SPIN.getId());
+        query.setInt(3, HIGH_ALTITUDE_SPIN.getId());
+
```
travisdesell, line 66: also fix/update the event statistics in here.

travisdesell (changes requested):

travisdesell (approved):

PR #45, opened 2022-09-09
Aidan/spin events

aidanlabella:


src/main/java/org/ngafid/FindSpinEvents.java
```diff
@@ -266,6 +272,15 @@ public static void main(String [] args) {
 
             LOG.info("Processing spin events for fleet: " + fleetId);
 
+            try {
```
travisdesell, line 275: Can remove this because those flights should already have been imported and things been calculated correctly.

travisdesell (changes requested):

src/main/java/org/ngafid/FindSpinEvents.java
```diff
@@ -28,7 +28,7 @@ public class FindSpinEvents {
 
     public static void findSpinEventsInUpload(Upload upload) {
         try {
-            String whereClause = "upload_id = " + upload.getId() + " AND insert_completed = 1 AND NOT EXISTS (SELECT flight_id FROM events WHERE id IN (" + HIGH_ALTITUDE_SPIN.getId() + ", " + LOW_ALTITUDE_SPIN.getId() + "))";
+            String whereClause = "upload_id = " + upload.getId() + " AND insert_completed = 1 AND NOT EXISTS (SELECT flight_id FROM events WHERE event_definition_id IN (" + HIGH_ALTITUDE_SPIN.getId() + ", " + LOW_ALTITUDE_SPIN.getId() + "))";
```
travisdesell, line 31: This is wrong for the same reason as Aaron's code. Look at line 598 at CalculateProximity.java

travisdesell (changes requested):

travisdesell (approved):

PR #57, opened 2022-10-14
LOCI Annotations List

AarC10:


aidanlabella (changes requested):
Unstage the *-bundle.js files. These are compiled by `npm` locally and don't need to be stored in the repo (they just take up extra space).

aidanlabella (approved):

PR #63, opened 2022-12-15
Add checks to make sure where aren't sending emails to null addresses that will throw exceptions.

aidanlabella:
Also made the email module create a default email info txt file if none exists.


src/main/java/org/ngafid/SendEmail.java
```diff
@@ -39,15 +45,41 @@ public class SendEmail {
...
+                if (username.startsWith("#")) {
+                    LOG.severe("Email not being used with the NGAFID for uploads. To change this, add the email login information to " + NGAFID_EMAIL_INFO);
+                } else {
+                    password = bufferedReader.readLine();
+                    LOG.info("Using email address to send emails: " + username);
+                }
+
+                //Don't remove this!
+                bufferedReader.close();
+            }
```
AarC10, line 81: Why not just do a try-with resources inside the else clause

AarC10 (approved):

travisdesell (approved):

PR #64, opened 2022-12-16
Refactor some code to make calculating things easier and more organized

aidanlabella:


src/main/java/org/ngafid/flights/calculations/HeadingTrackDiff.java
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                }
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.exit(1);
+        } 
+
+        LOG.info("All done!");
+        System.exit(0);
```
AarC10, line 124: Is the sys exit necessary here?

AarC10 (approved):

src/main/java/org/ngafid/flights/calculations/HeadingTrackDiff.java
aidanlabella: I find that it can be. Sometimes the maven process doesn't exit immediately on its own, so the script can hang for a bit. This just expedites the termination of the process so the OS can worry about other things (and so that there isn't a locked process in your terminal 🙂).

travisdesell (approved):

PR #65, opened 2022-12-27
DJI Binary Import

AarC10:


src/main/java/org/ngafid/ProcessUpload.java
```diff
@@ -352,6 +394,20 @@ public static boolean ingestFlights(Connection connection, int uploadId, int fle
                 UploadError.insertError(connection, uploadId, "Could not read from zip file: please delete this upload and re-upload.");
                 status = "ERROR";
                 uploadException = new Exception(e.toString() + ", could not read from zip file: please delete this upload and re-upload.");
+            } catch (NotDatFile e) {
+                System.err.println("NotDatFile: " + e);
+                e.printStackTrace();
+
+                UploadError.insertError(connection, uploadId, "Tried to process a non-DAT file as a DAT file.");
+                status = "ERROR";
```
aidanlabella, line 402: Consider making ERROR a const. static

src/main/java/org/ngafid/ProcessUpload.java
```diff
@@ -433,4 +489,39 @@ public static boolean ingestFlights(Connection connection, int uploadId, int fle
         //ingestion was successfull
         return true;
     }
+
+    private static void placeInZip(String file, String zipFileName) throws IOException {
+        System.out.println("Placing " + file + " in zip");
```
aidanlabella, line 494: I would advise using the `LOG` here for sysouts. Makes the output more uniform and traceable.

aidanlabella (commented):
Looks good! Just a few minor stylistic things for consistency wouldn't hurt 🙂

src/main/java/org/ngafid/ProcessUpload.java
AarC10: I will do this, but the entire ProcessUpload file uses Sysouts instead of a Logger which was the reasoning behind this

aidanlabella (approved):

PR #68, opened 2023-01-26
Surround Cesium Buttons on Flight Page With a Cell

AarC10:
Screenshot_20230126_132657

### DIFF
--- a/src/main/javascript/cesium_buttons.js
+++ b/src/main/javascript/cesium_buttons.js
@@ -27,7 +27,7 @@ class CesiumButtons extends React.Component {
         window.open(URL);
     }
 
-    createButtonStyle(color) {
+    createButtonStyle() {
         console.log("Creating style")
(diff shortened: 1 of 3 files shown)


aidanlabella:
just being a bit picky - I would justify the buttons to one side, add some spacing between the buttons and add a duplicate of the buttons to the bottom (like on the flights page).

aidanlabella:
and maybe also try to use the same color scheme as the paginator?

AarC10:
Screenshot_20230126_205151

Screenshot_20230126_205250

Did all your suggestions. Here is a screenshot comparison of both having just the top set of buttons and both the top and bottom. Just seems a little strange and might feel a little cluttering to me since I don't think we would expect a long list of flights on the flight page to necessitate a bottom one, but let me know what you think.

travisdesell (approved):

PR #70, opened 2023-02-13
[WIP] Cleaning Up

AarC10:
- More null pointer checks
- String builder and appends to create strings faster
- Reduced redundant if statements 
- Automatic resource closing for safety

### DIFF
--- a/src/main/java/org/ngafid/CalculateExceedences.java
+++ b/src/main/java/org/ngafid/CalculateExceedences.java
@@ -67,10 +67,12 @@ public class CalculateExceedences {
             System.out.println("minMaxRPM1: " + minMaxRPM1);
             System.out.println("minMaxRPM2: " + minMaxRPM2);
 
-            if ((minMaxRPM1 == null && minMaxRPM2 == null)  //both RPM values are null, can't calculate exceedence
-                    || (minMaxRPM2 == null && minMaxRPM1 != null && minMaxRPM1.second() < 800) //RPM2 is null, RPM1 is < 800
-                    || (minMaxRPM1 == null && minMaxRPM2 != null && minMaxRPM2.second() < 800) //RPM1 is null, RPM2 is < 800
-                    || (minMaxRPM1 != null && minMaxRPM1.second() < 800) && (minMaxRPM2 != null && minMaxRPM2.second() < 800)) { //RPM1 and RPM2 < 800
+            //both RPM values are null, can't calculate exceedence
+            //RPM2 is null, RPM1 is < 800
+            //RPM1 is null, RPM2 is < 800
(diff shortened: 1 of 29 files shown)


src/main/java/org/ngafid/CalculateExceedences.java
```diff
@@ -67,10 +67,10 @@ public void processFlight(Connection connection, Flight flight, EventDefinition
...
             System.out.println("minMaxRPM2: " + minMaxRPM2);
 
-            if ((minMaxRPM1 == null && minMaxRPM2 == null)  //both RPM values are null, can't calculate exceedence
-                    || (minMaxRPM2 == null && minMaxRPM1 != null && minMaxRPM1.second() < 800) //RPM2 is null, RPM1 is < 800
-                    || (minMaxRPM1 == null && minMaxRPM2 != null && minMaxRPM2.second() < 800) //RPM1 is null, RPM2 is < 800
-                    || (minMaxRPM1 != null && minMaxRPM1.second() < 800) && (minMaxRPM2 != null && minMaxRPM2.second() < 800)) { //RPM1 and RPM2 < 800
+            //both RPM values are null, can't calculate exceedence
+            //RPM2 is null, RPM1 is < 800
+            //RPM1 is null, RPM2 is < 800
+            if (minMaxRPM1 == null && minMaxRPM2 == null || minMaxRPM2 == null && minMaxRPM1.second() < 800 || minMaxRPM1 == null && minMaxRPM2.second() < 800 || minMaxRPM1 != null && minMaxRPM1.second() < 800 && minMaxRPM2 != null && minMaxRPM2.second() < 800) { //RPM1 and RPM2 < 800
```
travisdesell, line 73: please revert, those parenthesis mean things.

src/main/java/org/ngafid/GenerateBulkCSVS.java
```diff
@@ -350,7 +350,7 @@ public static void main(String[] args) {
                         }
 
 						sb.append(args[j]);
-						if (notEnd) sb.append(" ");
+						sb.append(" ");
```
travisdesell, line 353: I think we want to keep this to not have trailing whitespace.

src/main/java/org/ngafid/GenerateBulkCSVS.java
```diff
@@ -383,7 +383,7 @@ public static void main(String[] args) {
 			System.exit(1);
 		}
 
-		if (!dir.substring(dir.length() - 1, dir.length()).equals("/")) {
+		if (dir.charAt(dir.length() - 1) != '/') {
```
travisdesell, line 386: revert this too - instead it would be better to throw an exception if the directory length is 0

PR #72, opened 2023-02-20
Travis/remove gaard

travisdesell:
Removes references to GAARD from the front page as this has gone away.


aidanlabella (approved):
Looks good! I've been meaning to bring up removing GAARD references for a while now 😛 .

AarC10 (approved):

PR #81, opened 2023-03-29
Cleanup root directory

jkarns275:
Goal of this is to remove unnecessary clutter in the root directory, namely:

- Merging doc folders (NGAFID Web Design, schematics, merged into docs)
- Moving run scripts into a folder (`run` folder)


AarC10:
No complaints from me. I think it is fine

aidanlabella:
> Now that I think about it, do we need a few of those scripts still? For example, is there a purpose to having process_flights if we now use process_upload?

Yes, there is a purpose for having them because it isolates the flight ingestion stage in the data ingestion pipeline. This can be useful for debugging when we don't want to calculate exceedences and just import the CSV data.

jkarns275:
> Yes, there is a purpose for having them because it isolates the flight ingestion stage in the data ingestion pipeline. This can be useful for debugging when we don't want to calculate exceedences and just import the CSV data.

I agree but I think this functionality should probably just be added to the ProcessUpload.java file. It should serve as both a daemon and a CLI tool for development I think. Especially since there is a lot of code duplication between the two.

PR #84, opened 2023-04-10
Calculate rate of closure for proximity events

mayurjs26:


src/main/java/org/ngafid/CalculateProximity.java
```diff
@@ -2,46 +2,34 @@
...
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-
-
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.LinkedHashSet;
+import java.util.*;
```
travisdesell, line 21: Could you be specific here with the imports? It's not good form to do import * (it's slower and less specific).

db/create_tables.php
```diff
@@ -597,6 +598,19 @@
...
 
+if ($update_rate_of_closure) {
+    $query = "CREATE TABLE `rate_of_closure` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `event_id` INT(11) NOT NULL,
+    `size` INT(11) NOT NULL,
+    `data` MEDIUMBLOB,
+    PRIMARY KEY(`id`),
+    FOREIGN KEY(`event_id`) REFERENCES events(`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1";
```
travisdesell, line 609: Can you fix this formatting? Your indentation is bad.

src/main/java/org/ngafid/events/RateOfClosure.java
```diff
@@ -0,0 +1,87 @@
+package org.ngafid.events;
+
+import org.ngafid.common.Compression;
+
+import javax.sql.rowset.serial.SerialBlob;
+import java.io.IOException;
+import java.sql.*;
```
travisdesell, line 7: Could you expand this one out too?

travisdesell (changes requested):

travisdesell (approved):

PR #87, opened 2023-04-25
New Flight File Processing

AarC10:


src/main/java/org/ngafid/ProcessUpload.java
```diff
@@ -51,9 +50,9 @@ public class ProcessUpload {
     private static final Map<String, FileProcessor> fileProcessors = new HashMap<>();
 
     static {
-        fileProcessors.put(".csv", null);
-        fileProcessors.put(".json", null);
-        fileProcessors.put(".gpx", null);
+        fileProcessors.put(".csv", new CSVFileProcessor());
```
jkarns275: We can probably just change this to:

`Map.of(".csv", new CSVFileProcessor(), ...)`

so the declaration and definition can be together.

src/main/java/org/ngafid/ProcessUpload.java
```diff
@@ -298,25 +302,25 @@ public static boolean ingestFlights(Connection connection, int uploadId, int fle
         HashMap<String, UploadException> flightErrors = new HashMap<String, UploadException>();
 
         int validFlights = 0;
-        int warningFlights = 0;
-        int errorFlights = 0;
+        AtomicInteger warningFlights = new AtomicInteger();
```
jkarns275, line 305: I don't think these need to be atomic since we are not using a parallel stream. Though, we may want to use 'forEachOrdered' instead of `forEach` to make this more explicit.

src/main/java/org/ngafid/ProcessUpload.java
```diff
@@ -51,13 +44,24 @@
 import org.ngafid.accounts.Fleet;
 import org.ngafid.accounts.User;
 
-import static org.ngafid.flights.DJIFlightProcessor.processDATFile;
+
+@FunctionalInterface
+interface FlightFileProcessors {
```
jkarns275, line 49: Not in love with the `Object...`. We can probably do this in a cleaner way.

jkarns275: The move is probably going to be to move to using `FlightFileProcessor::create`, which takes in the `ZipEntry` + creates the `InputStream`.

src/main/java/org/ngafid/flights/process/FlightMeta.java
```diff
@@ -19,4 +19,108 @@ public final class FlightMeta {
                   airframeName,
                   calculated,
                   suggestedTailNumber;
+
```
jkarns275, line 22: Kotlin be like

jkarns275 (approved):
I think it looks really good. If you want to merge it into the staging branch I can try and polish the places where our code connects.

PR #112, opened 2023-07-21
Fixing import for JSON

KanishaAgrawal:


travisdesell (changes requested):

src/main/java/org/ngafid/flights/Flight.java
```diff
@@ -2755,13 +2806,11 @@ public static <T> Flight processJSON(int fleetId, Connection connection, InputSt
...
                        
                         if ((line.get(i) instanceof Double) || doubleSeries.containsKey(column)) {
-                            
-                            String unit = doubleUnitMap.getOrDefault(column, "unknown");
                             DoubleTimeSeries dseries = doubleSeries.get(column);
                            
                             if (dseries == null) {
                               
-                               dseries = new DoubleTimeSeries(connection, column, unit, len);
+                               dseries = new DoubleTimeSeries(connection, column, dataType, len);
```
jkarns275, line 2813: This entire switch can be refactored. All of these special cases can be handled in their own break out functions.

Basically, just get all of the data into the time series and then call functions to process the data after the fact e.g. 
```
    for (ArrayList<T> line : lines) {
        for (int i = 0; i < line.size(); i++) {
            String column = columns.get(i);
            if (doubleSeries.containsKey(column)) {
                ...
            } else {
                ...
            }
        }
    }


    processAltAGL(doubleSeries.get("AltAGL"));
    processGndSpd(doubleSeries.get("GndSpd"));
    ...
```

jkarns275 (changes requested):

PR #116, opened 2023-09-14
Added system id and tail to the labels in the severity graph plot

e911:
Screen Shot 2023-09-14 at 11 10 29 AM

### DIFF
--- a/src/main/java/org/ngafid/events/Event.java
+++ b/src/main/java/org/ngafid/events/Event.java
@@ -35,6 +35,10 @@ public class Event {
 
     private Integer otherFlightId = null;
 
+    private String systemId;
+
+    private String tail;
+
     private RateOfClosure rateOfClosure;
 
     public Event(String startTime, String endTime, int startLine, int endLine, double severity) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/ngafid/events/Event.java
```diff
@@ -35,6 +35,10 @@ public class Event {
 
     private Integer otherFlightId = null;
 
+    private String systemId;
+
+    private String tail;
```
AarC10, line 40: Curious if we can make things final here.

src/main/java/org/ngafid/events/Event.java
```diff
@@ -35,6 +35,10 @@ public class Event {
...
+
     private RateOfClosure rateOfClosure;
 
     public Event(String startTime, String endTime, int startLine, int endLine, double severity) {
```
AarC10, line 44: Is this constructor still used anywhere else? Would using this constructor break something if there is no system ID or tail number provided?

e911: Yeah, this constructor was used on other files and I had to make a new one just to accomodate for adding the system and tail id.

PR #120, opened 2023-10-25
New flight file format

e911:


src/main/java/org/ngafid/flights/Flight.java
```diff
@@ -1796,12 +1806,16 @@ private void initialize(Connection connection, InputStream inputStream) throws F
 
                 airframeName = "ScanEagle";
                 airframeType = "UAS Fixed Wing";
+            } else if(fileInformation.startsWith("Aircraft ID")) {
+                airframeName = "N709EA";
```
travisdesell, line 1810: The airframe name is actually `Beechcraft C90A King Air`.  N709EA is the tail number. We can set both the tail number and system ID to 'N709EA'.

travisdesell (changes requested):

travisdesell (approved):

PR #143, opened 2024-06-07
Aidan cu/email preferences

Ilverism:
A new system for managing various types of Email Preferences, so Users can enable or disable different types of emails. Managers can also toggle these settings on behalf of their fleet users.

- New table `email_preferences` in `create_tables.php`
- New email definition + compilation script: `EmailTypes.java` / `generate_email_types.sh`
- New `GET` / `POST` routes, `email_preferences` and `update_email_preferences`
- New UI table for user email type handling
- New UI table for manager email type handling

---

USAGE GUIDE _(copy of docs/EmailPreferencesGuide_V1.md)_

**Email Preferences:**

- <span style="text - decoration: underline;">Back-End Setup</span>

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;1. Create database table with `create_tables.php`

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;2. Define/Configure Email Types in `EmailTypes.java`

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;3. Generate Email Types with `generate_email_types.sh`

- <span style="text - decoration: underline;">Front-End Usage</span>
\
&ensp;&ensp;**○** Standard Users:
\
&ensp;&ensp;&ensp;&ensp;**🞄** Configure their own email preferences inside **Account → My Preferences**
\
\
&ensp;&ensp;**○** Fleet Managers:
\
&ensp;&ensp;&ensp;&ensp;**🞄** Configure their users’ preferences in **Account → Manage Fleet**
\
&ensp;&ensp;&ensp;&ensp;**🞄** Can configure their own preferences like a standard user
\
&ensp;&ensp;&ensp;&ensp;**🞄** Can mass-toggle their users’ preferences with the big checkboxes (☑️ / ⬜)
\
\
&ensp;&ensp;**○** General:
\
&ensp;&ensp;&ensp;&ensp;**🞄** Some Email Types are only visible to *admin* accounts
\
&ensp;&ensp;&ensp;&ensp;**🞄** Some Email Types are hidden to *all* accounts


src/main/java/org/ngafid/EmailType.java
```diff
@@ -0,0 +1,260 @@
+package org.ngafid;
```
jkarns275, line 1: Should probably be moved into org.ngafid.accounts

src/main/java/org/ngafid/SendEmail.java
```diff
@@ -23,125 +33,149 @@ public class SendEmail {
...
-        }
+            }
+
         String NGAFID_ADMIN_EMAILS = System.getenv("NGAFID_ADMIN_EMAILS");
         adminEmails = new ArrayList<String>(Arrays.asList(NGAFID_ADMIN_EMAILS.split(";")));
         System.out.println("import emails will always also be sent to the following admin emails:");
         for (String adminEmail : adminEmails) {
             System.out.println("\t'" + adminEmail + "'");
-        }
+            }
```
jkarns275, line 66: how does this keep happening

src/main/java/org/ngafid/SendEmail.java
```diff
@@ -23,125 +33,149 @@ public class SendEmail {
...
 
         try {
+
             File file = new File(NGAFID_EMAIL_INFO);
             BufferedReader bufferedReader = null;
             
-            if (!file.exists()) {
-                // Create a new file and will populate it 
+            //Email info file does not exit...
+            if (file.exists() == false) {
```
jkarns275, line 74: boolean constants should not usually be directly compared to

src/main/java/org/ngafid/SendEmail.java
```diff
@@ -177,16 +211,30 @@ public static void sendEmail(ArrayList<String> toRecipients, ArrayList<String> b
 
                 // Set To: header field of the header.
                 for (String toRecipient : toRecipients) {
+
                     //list of users who do not want emails: TODO: make this a user setting
-                    if (toRecipient.equals("nievesn2@erau.edu")) continue;
+                    if (toRecipient.equals("nievesn2@erau.edu"))
```
jkarns275, line 216: Once we land this, we can manually disable these settings for this user (as a courtesy) and then get rid of this.

src/main/java/org/ngafid/SendEmail.java
```diff
@@ -177,16 +211,30 @@ public static void sendEmail(ArrayList<String> toRecipients, ArrayList<String> b
...
+                        continue;
+
+                    //Check if the emailType is forced
+                    if (EmailType.isForced(emailType))
+                        System.out.println("Delivering FORCED email type: " + emailType);
+
+                    //Check whether or not the emailType is enabled for the user
+                    else if (UserPreferencesEmails.getEmailTypeUserState(toRecipient, emailType) == false)
```
jkarns275, line 224: use negation

src/main/java/org/ngafid/accounts/User.java
```diff
@@ -35,6 +39,8 @@ public class User {
     private boolean admin;
     private boolean aggregateView;
 
+    private UserPreferencesEmails userPreferencesEmails;
```
jkarns275, line 42: `UserEmailPreferences`

src/main/java/org/ngafid/accounts/User.java
```diff
@@ -227,6 +239,10 @@ public static User get(Connection connection, int userId, int fleetId) throws SQ
         user.fleetAccess = FleetAccess.get(connection, user.id, fleetId);
         user.fleet = Fleet.get(connection, fleetId);
 
+        //Get the email preferences for this user
+        user.userPreferencesEmails = getUserPreferencesEmails(connection, userId);
```
jkarns275, line 243: Make this lazy, since we won't need the email preferences 99% of the time. Just set to null by default and the getter function can populate it.

src/main/java/org/ngafid/accounts/UserPreferencesEmails.java
```diff
@@ -0,0 +1,131 @@
...
+
+
+public class UserPreferencesEmails {
+
+    private int userId;
+    private HashMap<String, Boolean> emailTypesUser;
+    private String[] emailTypesKeys;
+
+    private static HashMap<Integer, User> users = new HashMap<Integer, User>();
+    private static HashMap<String, HashMap<String, Boolean>> emailTypesUsers = new HashMap<String, HashMap<String, Boolean>>();
```
jkarns275, line 29: I don't know if we should cache these things since they could be changed at any time by the web server?

jkarns275 (commented):
Code needs to be formatted to match the rest of the codebase, other than that just minor revisions.

jkarns275 (changes requested):
Gating

src/main/java/org/ngafid/accounts/EmailType.java
```diff
@@ -98,6 +101,16 @@ public static EmailType[] getAllTypes() {
         return values();
     }
 
+    public static String[] getEmailTypeKeysRecent(Boolean doRefresh) {
```
jkarns275, line 104: In java `boolean` is the primitive boolean type and `Boolean` is a boxed version (i.e. it is heap allocated as an object which contains a `boolean`)

src/main/java/org/ngafid/accounts/EmailType.java
```diff
@@ -167,13 +188,14 @@ public static void insertEmailTypesIntoDatabase(int userIDTarget) {
...
         }
 
-
         StringBuilder query = new StringBuilder();
         query.append("INSERT INTO email_preferences (user_id, email_type) ");
 
         //Generate "individual" email type queries
         for (EmailType emailType : EmailType.values()) {
-            query.append("SELECT ").append(userIDTarget).append(", '").append(emailType.getType()).append("' UNION ALL ");
+            String emailTypeKey = emailType.getType();
```
jkarns275, line 196: Can probably replace this with `String::join`, which will join a list of strings with some delimiter. In this case `UNION ALL` will be your delimiter. This pattern comes up frequently enough that they made a utility function for it ;)

src/main/java/org/ngafid/accounts/EmailType.java
```diff
@@ -167,13 +188,14 @@ public static void insertEmailTypesIntoDatabase(int userIDTarget) {
...
+            emailTypeKeysRecent.add(emailTypeKey);
+            query.append("SELECT ").append(userIDTarget).append(", '").append(emailTypeKey).append("' UNION ALL ");
         }
 
         //Remove the last "UNION ALL" from the query
```
jkarns275, line 201: Remove prints. If you have logging information you want to keep in production for our logs, use the logger.

src/main/java/org/ngafid/accounts/User.java
```diff
@@ -87,8 +87,15 @@ public boolean hasAggregateView() {
...
-        return userEmailPreferences;
+    public UserEmailPreferences getUserEmailPreferences(Connection connection) {
+
+        try {
+            return User.getUserEmailPreferences(connection, id);
+        } catch (SQLException e) {
+            e.printStackTrace();
+            return null;
+        }
+
```
jkarns275, line 98: This function should probably be declared to `throw SLQException` so it can propagate the error up to our other error handling code. Otherwise, we will just get a null pointer exception.

src/main/java/org/ngafid/accounts/User.java
```diff
@@ -240,7 +247,7 @@ public static User get(Connection connection, int userId, int fleetId) throws SQ
         user.fleet = Fleet.get(connection, fleetId);
 
         //Get the email preferences for this user
-        user.userEmailPreferences = getUserEmailPreferences(connection, userId);
+        //user.userEmailPreferences = getUserEmailPreferences(connection, userId);
         UserEmailPreferences.addUser(user);
```
jkarns275, line 251: This method requires a connection, you should just pass it as a parameter. Not sure if it is "best practice" but its usually what we do in the rest of the codebase. Look around and you'll see a lot of methods with `Connection connection` as the first parameter.

src/main/java/org/ngafid/routes/GetUserEmailPreferences.java
```diff
@@ -45,38 +56,27 @@ public Object handle(Request request, Response response) {
...
-                response.status(400);
-                return gson.toJson(new ErrorResponse("[GET Email Preference Error]", "Specified an invalid userID parameter"));
-            }
+        //Fetching Session User...
+        if (handleFetchType.equals("HANDLE_FETCH_USER")) {
+            fleetUserID = sessionUser.getId();
+        }
 
+        //Fetching a Manager's Fleet User...
+        else if (handleFetchType.equals("HANDLE_FETCH_MANAGER")) {
```
jkarns275, line 74: Validate that the current use has these permissions / is a manager

src/main/javascript/email_settings.js
```diff
@@ -33,6 +33,7 @@ const EmailSettingsTableUser = ({ isAdmin }) => {
             url: '/protected/update_email_preferences',
             data: submissionData,
             dataType: 'json',
+            async: false,
```
jkarns275, line 36: Is this to make sure the buttons don't get fiddled with until the preferences have been updated in the backend?

src/main/javascript/email_settings.js
```diff
@@ -51,15 +52,21 @@ const EmailSettingsTableUser = ({ isAdmin }) => {
     function getUserEmailPreferences() {
 
         let resultsOut = "No results found.";
+
+        let submissionData = {
+            handleFetchType : "HANDLE_FETCH_USER"
+            }
```
jkarns275, line 58: <----

src/main/javascript/email_settings.js
```diff
@@ -276,6 +283,7 @@ const EmailSettingsTableManager = ({ fleetUsers }) => {
             url: '/protected/update_email_preferences',
             data: submissionData,
             dataType: 'json',
+            async: false,
```
jkarns275, line 286: So just make sure the async flags are set intentionally. These freeze the web page so we don't want to do that unless its necessary. Usually you would disable some elements on the web page until the contingent request returns rather than using a synchronous call.

www_php/js/uploads.js
jkarns275, line 1: This directory basically needs to be removed. Our javascript is in src/main/javascript.

jkarns275 (changes requested):

src/main/java/org/ngafid/routes/GetUserEmailPreferences.java
jkarns275: I don't think this was addressed -- you will have to take the user id of the current user and validate that they are in fact a manager.

src/main/java/org/ngafid/routes/GetUserEmailPreferences.java
```diff
@@ -0,0 +1,90 @@
...
+        User sessionUser = session.attribute("user");
+
+
+        int fleetUserID = -1;
+        
+        //Fetching Session User...
+        if (handleFetchType.equals("HANDLE_FETCH_USER")) {
+            fleetUserID = sessionUser.getId();
+        }
+
```
jkarns275, line 72: Usually we wrap if statements e.g.
```
if (x) {
  ...
} else if (y) {
  ...
} else {
  ...
}
```

The gap between the if and the else is sort of misleading to most people

travisdesell (approved):

PR #144, opened 2024-06-19
Unsubcribe Links for Email Preferences

Ilverism:
Adds new links to the top of received emails which, when clicked, will disable all non-forced email types in the database.

Consequently, all standard recipients will have separate emails delivered to support the unique unsubscribe links, as opposed to having them batched like before.

The BCC recipients will get a separate email which does not contain an unsubscribe link.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ db/db_info.php
 db/db_info
 email_info.txt
 package-lock.json
+data/archive/*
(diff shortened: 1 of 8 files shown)


jkarns275:
This is pretty much perfect to roll out.

In a follow up PR, I think the unsubscribe links should be per-email type? I'm thinking you have two links: "no more emails like this" and "unsubscribe from all emails." 

Then another link to the ngafid website with "manage email preferences" to the preferences page.

Lets wait until tomorrow to merge though.

PR #147, opened 2024-06-20
Fix empty default severity columns

Ilverism:
Records the default value of the Severity Columns so empty values can no longer be added, preventing a potential NullPointerException when managing events or viewing event statistics


jkarns275:
It looks good but I don't know a lot about the code surrounding this

PR #149, opened 2024-06-26
Flights and Flight page bugfixes/changes

Ilverism:
Changes and bugfixes for Flights / Flight pages:

- Removed horizontal scrollbar at the bottom of the page
- Moved the lower Filter scrollbar to the left side of the page
- Changed the lower Filter scrollbar behavior to auto
- Added padding to the bottom of the Filter div so the entire area can be viewed without having to scroll down all the way with the larger scrollbar
- Changed Saved Filters scrolling to auto
- Changed Add a Tag scrolling to auto
- Shortened the length of the obtrusive 10 second error message popup
- Swapped the display position of the plot and map to match the order of the buttons at the top
- Renamed “Store Filter” to “Save Filter”
- Prevented the saving of an empty/invalid filter after opening the save modal and clearing/changing the rules
- Truncated the “Sorting By” dropdown text with an ellipsis to reduce the chance of items being rendered across multiple lines
- The “Go To” button no longer resizes if items are rendered across multiple lines
- Disabled user text selection for the page change buttons
- Disabled text selection on a space between the view/clear replay buttons
- All of the flight button sizes now match
- All of the header button sizes now match
- All of the tag button sizes now match
- Prevented the map from filling the screen when toggling the plot off
- The X danger buttons now activate when clicking anywhere on the button instead of just the X icon
- Made the order of displayed information match the button order  (Events → Tags → Parameters → Itinerary)
- Removed the Add a Tag + button and just made it so the drop down menu for creating/adding tags is always visible
- Added a message box indicating if no events were found for a flight, similar to the existing one that indicates if there are no tags for a flight
- Fixed the incorrect (?) ‘title=globeTooltip’ and globeClasses’ in the flight map button
- The page number now resets when clicking the Apply Filter button
- If there are no filter values available (e.g. for airport/runway/tag rules), this is now indicated by disabling the drop down menu and inserting “N/A”
- Imported Error and Confirm modals to flight.js
- Improved the functionality/reliability of the tag selection and removal


Ilverism:
Outdated.

Newer version part of https://github.com/travisdesell/ngafid2.0/pull/152

PR #154, opened 2024-08-16
Fix button access on Uploads page for users with View/Upload privileges

Ilverism:
Fixed/updated the behavior for the following on the Uploads page:

**VIEW ACCESS:**
```
* Upload Button   -> Disabled ❌
* Delete Button   -> Disabled ❌
* Download Button -> Enabled  ✅
```

**UPLOAD & MANAGER ACCESS:**
```
* Upload Button   -> Enabled  ✅
* Delete Button   -> Enabled  ✅
* Download Button -> Enabled  ✅
```

08/26/24:
- The Upload/Download buttons will now be completely hidden for users with View access. Functionality to disable/gray-out certain buttons is kept for when the delete and download buttons during an in-progress upload, and for posterity for the upload button.
- Now uses information injected by Navbar's getJavascript to determine if the Upload/Delete buttons can be displayed (existing fix merged from commit on main)

### DIFF
--- a/src/main/java/org/ngafid/routes/GetUpload.java
+++ b/src/main/java/org/ngafid/routes/GetUpload.java
@@ -44,9 +44,10 @@ public class GetUpload implements Route {
             return "Upload not found";
         }
 
-        if (!user.hasUploadAccess(upload.getFleetId())) {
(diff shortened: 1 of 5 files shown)


jkarns275 (approved):
LGTM

PR #158, opened 2024-09-09
Split CSV files based on time gap indicating new flight.

eightspokes:
**Feature:** 
This is a Draft PR (Proof of Concept) for the File Splitter feature to segregate flights if the time gap between entries is too large ( currently set to 5 minutes). 

**Integration testing:**
C175 file was modified to emulate large time gap. 
After Process Upload runs, two separate CSV files are created and placed into a Zip file for further processing. 
The Webpage correctly displays two flights from a single C175 CSV file.

**To DO:** 
Finalize the solution with the team
Add logs
Cleanup
Testing

### DIFF
--- a/src/main/java/org/ngafid/ClearUpload.java
+++ b/src/main/java/org/ngafid/ClearUpload.java
@@ -11,14 +11,12 @@ public class ClearUpload {
     private static Connection connection = null;
     private static Logger LOG = Logger.getLogger(ProcessUpload.class.getName());
     private static final String ERROR_STATUS_STR = "ERROR";
-    
+
     public static void main(String[] arguments) {
         System.out.println("arguments are:");
         System.out.println(Arrays.toString(arguments));
 
-        try {
-            connection = Database.getConnection();
-
+        try (Connection connection = Database.getConnection()) {
             int uploadId = Integer.parseInt(arguments[0]);
             Upload.clearUpload(connection, uploadId);
         } catch (SQLException e) {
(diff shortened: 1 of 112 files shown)


jkarns275:
Code looks good, just had a few general thoughts:
- As we discussed in the meeting, we need a boolean column in the `uploads` table to mark derived uploads.
- We have some code that allows us to re-process a particular upload. In order to properly delete an upload that has derived uploads, we need to keep that information somewhere. Perhaps we should add a nullable field in the uploads table that points to the parent upload ID (or null for non-derived uploads).
- If we are re-processing an upload that has derived uploads, all flights and data associated with those uploads should be deleted and the derived uploads themselves should be deleted. The code to do the deleting of flights and data etc. is already in place, but deleting the upload and the file for derived flights would be new.
- As it stands, I think the code would still attempt to parse the broken CSV files in the original upload.
- We probably want to only allow 1 derived upload per upload, just for organizational purposes. 
- I'm not 100% sure of what the best way to do it is quite yet, but I think the CSV splitting should happen in the flight processing pipeline. I already have some code there for modifying the zip file in place, but as we discussed this will be changed to create a new derived upload.

jkarns275:
I am going to have to tweak the way the flight processing pipeline works a bit to accommodate a derived upload, but it should not be difficult. I will work on that soon so I don't delay you. 

What I have in mind will process all of the flights in the one upload, but the derived upload will still need to be established so users can still download flight files correctly.

PR #159, opened 2024-09-24
Spotless Formatter

AarC10:
Add a maven plugin for running a format check and autoformatter called [Spotless](https://github.com/diffplug/spotless).

Can be ran locally by doing 
`mvn spotless:check` which will display violating files
`mvn spotless:apply` will handle formatting the files for you

Should be merged after the backlog of other changes are merged into main and this branch


jkarns275:
After all pending PRs merge, I would recommend re-creating this branch on the new base.

jkarns275:
Very soon we should apply this. Most people seemt o be done with their big tasks / merged with main.

AarC10:
image
bro

jkarns275:
Thinking about this

PR #162, opened 2024-10-04
Liquibase database management tool

jkarns275:
[Liquibase](https://docs.liquibase.com/concepts/introduction-to-liquibase.html) is a database schema management tool.

Liquibase should make it easy for us to:
- Update and create the database
- Specify and update static values (e.g. events)

Summary is:
1. Database tables are defined in changelogs which contain "changesets",  (see: src/main/resources/changelogs).
2. These changelog files are loaded in alphanumeric order, and changesets within a file are loaded in top-to-bottom order.
3. Changesets are basically a sequence of SQL queries.
4. Changesets should not be changed: modification of a table should be done w/. a new changeset. Liquibase keeps track of this as well in the form of an md5 hash of the changeset code.
5. Each changeset is notated by some metadata in the form of a comment. Each changeset requires a unique id and an author name. You can add tags as well.
6. With a maven command, you can run all changesets / logs. If some have already been ran, only new changesets will be run. 

The new instructions for creating the database are (soon to be) in the README.md, as well as updating the database when new changesets are pulled in.


src/main/java/org/ngafid/Database.java
```diff
@@ -38,7 +37,7 @@ public static Connection getConnection() throws SQLException {
     }
 
     public static boolean dbInfoExists() {
-        return dbHost != null || dbName != null || dbUser != null || dbPassword != null;
+        return dbUrl != null && dbUser != null && dbPassword != null;
```
AarC10, line 40: Can we surround them with parentheses for readbaility

src/main/resources/changelogs/00-accounts/02-airsync.sql
```diff
@@ -0,0 +1,28 @@
+--liquibase formatted sql
+
+-- changeset josh:airsync labels:accounts,airsync
```
AarC10, line 3: Should they have your name on them?

src/main/resources/changelogs/01-flights/06-event-definitions.sql
```diff
@@ -0,0 +1,60 @@
+--liquibase formatted sql
```
AarC10, line 1: Sinec this only gets ran once when setting up, why not break up the insert statement by event definition type for readability? i.e group Total Fuel together as one insert statement

AarC10 (changes requested):

src/main/resources/changelogs/01-flights/06-event-definitions.sql
jkarns275: It would probably be good to do. I originally just copy and pasted this from the event definition blobs.

travisdesell: we need to make sure these are the most up to date ones from the database (not some other location). i did go in to fix a couple bugs with a few of these not too long ago (related to CHT and EGT events).

jkarns275: Will do this in a follow up PR before final deployment

PR #163, opened 2024-10-13
Roman/file split feature2

eightspokes:
**PR Objective:**
The CSV data from G3X and G5 flight recorders is inconsistent with our current parsing logic. This pull request addresses the changes needed to process G3X and G5 data and introduces functionality to split CSV entries into separate flights when there is a time gap larger than 5 minutes.

This PR also contains changes to the javascript files responsible for managing fleet. The users whose access level is denied are moved to the bottom of the list and grayed out. Also email section is omitted for such users.

**Changes:**
- In CSVFileProcessor.java modified the parse method to handle G3X and G5 CSV data. 
- Added the isG5FlightRecorder method to detect G5 files.
- Introduced the splitCSVIntoFlights method to split entries into separate flights based on time gaps
- G5 files do not contain the system.id value, so the serial_number is now used as a system_id (with a mapping implemented).

Fleet management ui changes:
- modified manage_fleet.js, email_settings.js

**_Airframe Name Handling:_**
G3X and G5 files do not include airframe_name. For testing, airframe_name is temporarily hardcoded to "Cessna 172" until Vig's changes are integrated.

**_File Creation:_**
When new derived flights are created, corresponding CSV files are generated and archived in a zip folder under the archive/derived directory.

**_Date and Time Adjustments:_**
G5 files do not contain local date/time values or UTC offsets.
Added the calculateLocalDateTime method in TimeUtils to compute local date/time and UTC offset from universal time and coordinates. This utilizes an external library for time zone conversion:
https://github.com/dustin-johnson/timezonemap

**Development Testing:**
For the G5 data that contain two flights.
Two flights are created. 
Screenshot 2024-10-12 at 8 33 02 PM
two derived csv files are created in a new zip folder:
Screenshot 2024-10-12 at 8 07 00 PM

Derived CSV files contain original headers and entries that correspond to the derived flights. 
Screenshot 2024-10-12 at 8 34 57 PM

Showing DENIED users at the bottom of the list, removing email section for DENIED users. 
Added a button to hide/reveal denied users.
Screenshot 2024-11-18 at 5 43 52 PM

### DIFF
--- a/src/main/java/org/ngafid/common/TimeUtils.java
+++ b/src/main/java/org/ngafid/common/TimeUtils.java
@@ -1,5 +1,4 @@
 package org.ngafid.common;
-
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZonedDateTime;
@@ -7,11 +6,11 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoUnit;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.List;
+import java.util.*;
 
+import us.dustinj.timezonemap.TimeZoneMap;
+import java.time.ZoneId;
+import java.util.Optional;
 import org.ngafid.flights.FatalFlightFileException;
 
 public class TimeUtils {
@@ -283,5 +282,94 @@ public class TimeUtils {
         cal.setTime(date);
         cal.add(Calendar.MILLISECOND, milliseconds);
         return cal.getTime();
+    }
+
+    /**
+     * Calculates local date, local time, and UTC offset for each entry based on given UTC dates, times, latitudes, and longitudes.
+     * @param map          map object (created once per upload for efficiency.
+     * @param utcDates     list of UTC dates as strings (e.g., "yyyy-MM-dd")
+     * @param utcTimes     list of UTC times as strings (e.g., "HH:mm:ss")
+     * @param latitudes    list of latitude strings
+     * @param longitudes   list of longitude strings
+     * @return a LocalDateTimeResult containing lists of local dates, times, and UTC offsets.
+     */
+    public static LocalDateTimeResult calculateLocalDateTime(TimeZoneMap map,
+            ArrayList<String> utcDates, ArrayList<String> utcTimes,
+            ArrayList<String> latitudes, ArrayList<String> longitudes) {
+
+
+
+        // Validate that all lists are of the same size.
+        if (utcDates.size() != utcTimes.size() || utcDates.size() != latitudes.size() || utcDates.size() != longitudes.size()) {
+            throw new IllegalArgumentException("All input lists must have the same size.");
+        }
+
+        // Prepare lists for the results.
+        ArrayList<String> localDates = new ArrayList<>(utcDates.size());
+        ArrayList<String> localTimes = new ArrayList<>(utcDates.size());
+        ArrayList<String> utcOffsets = new ArrayList<>(utcDates.size());
+
+        // Iterate over each index and calculate the corresponding local date, time, and offset.
+        for (int i = 0; i < utcDates.size(); i++) {
+            String dateTimeString = utcDates.get(i) + " " + utcTimes.get(i);
+            LocalDateTime utcDateTime = null;
+
+            // Try parsing using a formatter.
+            for (DateTimeFormatter formatter : dateTimeFormatters) {
+                try {
+                    utcDateTime = LocalDateTime.parse(dateTimeString, formatter);
+                } catch (Exception e) {
+                    // Continue trying other formats
+                }
+            }
+            // if utcTime has not been initialized still
+            if (utcDateTime == null) {throw new IllegalArgumentException("Invalid date/time format: " + dateTimeString);}
+
+            double latitude = Double.parseDouble(latitudes.get(i));
+            double longitude = Double.parseDouble(longitudes.get(i));
+
+
+            String zoneIdStr = map.getOverlappingTimeZone(latitude, longitude).getZoneId();
+            ZoneId zoneId = ZoneId.of(zoneIdStr);
+
+            // Convert the UTC datetime to the local time in the specified zone.
+            ZonedDateTime localZonedDateTime = utcDateTime.atZone(ZoneOffset.UTC)
+                    .withZoneSameInstant(zoneId);
+
+            // Format the local date, local time, and UTC offset.
+            localDates.add(localZonedDateTime.toLocalDate().format(DateTimeFormatter.ofPattern("yyyy-MM-dd")));
+            localTimes.add(localZonedDateTime.toLocalTime().format(DateTimeFormatter.ofPattern("HH:mm:ss")));
+            utcOffsets.add(localZonedDateTime.getOffset().getId());
+        }
+
+        return new LocalDateTimeResult(localDates, localTimes, utcOffsets);
+    }
+
+
+
+    /**
+     * Helper class to store the results of the date-time calculation.
+     */
+    public static class LocalDateTimeResult {
+        private final ArrayList<String> localDates;
+        private final ArrayList<String> localTimes;
+        private final ArrayList<String> utcOffsets;
+
+        public LocalDateTimeResult(ArrayList<String> localDates, ArrayList<String> localTimes, ArrayList<String> utcOffsets) {
+            this.localDates = localDates;
+            this.localTimes = localTimes;
+            this.utcOffsets = utcOffsets;
+        }
+        public ArrayList<String> getLocalDates() {
+            return localDates;
+        }
+
+        public ArrayList<String> getLocalTimes() {
+            return localTimes;
+        }
+
+        public ArrayList<String> getUtcOffsets() {
+            return utcOffsets;
+        }
     }
 }
--- a/src/main/java/org/ngafid/common/TimeUtils.java
+++ b/src/main/java/org/ngafid/common/TimeUtils.java
@@ -113,14 +113,39 @@ public class TimeUtils {
         return utcZdt.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"));
 
     }
+    /**
+     * List of DateTime formats supported.
+     */
+    private static final List<DateTimeFormatter> dateTimeFormatters = Arrays.asList(
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"),
+            DateTimeFormatter.ofPattern("M/d/yyyy HH:mm:ss")
+
+    );
+
+    /**
+     * Helper method to parse the date and time using the declared formatters
+     * @param dateTimeString
+     * @return
+     */
+    public static LocalDateTime parseDateTime(String dateTimeString) {
+        String normalizedDateTime = dateTimeString.replaceAll("\\s+", " ");  // Replaces multiple spaces with a single space
+        for (DateTimeFormatter formatter : dateTimeFormatters) {
+            try {
+                return LocalDateTime.parse(normalizedDateTime, formatter);
+            } catch (Exception e) {
+                // Continue trying other formats
+            }
+        }
+        // If none of the formats work, throw an exception or handle it appropriately
+        throw new IllegalArgumentException("Date format not supported: " + normalizedDateTime);
+    }
 
     public static OffsetDateTime convertToOffset(String originalDate, String originalTime, String originalOffset,
             String newOffset) {
         // System.out.println("original: \t" + originalTime + " " + originalOffset + " new offset: "+ newOffset);
 
         // create a LocalDateTime using the date time passed as parameter
-        LocalDateTime ldt = LocalDateTime.parse(originalDate + " " + originalTime,
-                DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"));
+        LocalDateTime ldt = parseDateTime(originalDate + " " + originalTime);
 
         // fix bad offset values
         originalOffset = updateBadOffset(ldt, originalOffset);
--- a/src/main/java/org/ngafid/common/TimeUtils.java
+++ b/src/main/java/org/ngafid/common/TimeUtils.java
@@ -8,6 +8,8 @@ import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
 
+import org.ngafid.flights.DoubleTimeSeries;
+import org.ngafid.flights.StringTimeSeries;
 import us.dustinj.timezonemap.TimeZoneMap;
 import java.time.ZoneId;
 import java.util.Optional;
@@ -281,7 +283,64 @@ public class TimeUtils {
         return cal.getTime();
     }
 
+
+    public static LocalDateTimeResult calculateLocalDateTimeFromTimeSeries(
+            TimeZoneMap map,
+            StringTimeSeries utcDates,
+            StringTimeSeries utcTimes,
+            DoubleTimeSeries latitudes,
+            DoubleTimeSeries longitudes) throws Exception {
+
+        // Validate that all time series are of the same size.
+        if (utcDates.size() != utcTimes.size() || utcDates.size() != latitudes.size() || utcDates.size() != longitudes.size()) {
+            throw new IllegalArgumentException("All input time series must have the same size.");
+        }
+
+        // Prepare lists for the results.
+        ArrayList<String> localDates = new ArrayList<>(utcDates.size());
+        ArrayList<String> localTimes = new ArrayList<>(utcDates.size());
+        ArrayList<String> utcOffsets = new ArrayList<>(utcDates.size());
+
+        // Iterate over each index and calculate the corresponding local date, time, and offset.
+        for (int i = 0; i < utcDates.size(); i++) {
+            String dateTimeString = utcDates.get(i) + " " + utcTimes.get(i);
+            LocalDateTime utcDateTime = null;
+
+            // Try parsing using a formatter.
+            for (DateTimeFormatter formatter : DATE_FORMATTERS) {
+                try {
+                    utcDateTime = LocalDateTime.parse(dateTimeString, formatter);
+                    break;
+                } catch (DateTimeParseException e) {
+                    // Continue trying other formats
+                }
+            }
+
+            if (utcDateTime == null) {
+                throw new IllegalArgumentException("Failed to parse date/time string: " + dateTimeString);
+            }
+
+            double latitude = latitudes.get(i);
+            double longitude = longitudes.get(i);
+
+            String zoneIdStr = map.getOverlappingTimeZone(latitude, longitude).getZoneId();
+            ZoneId zoneId = ZoneId.of(zoneIdStr);
+
+            // Convert the UTC datetime to the local time in the specified zone.
+            ZonedDateTime localZonedDateTime = utcDateTime.atZone(ZoneOffset.UTC).withZoneSameInstant(zoneId);
+
+            // Format the local date, local time, and UTC offset.
+            localDates.add(localZonedDateTime.toLocalDate().format(DateTimeFormatter.ofPattern("yyyy-MM-dd")));
+            localTimes.add(localZonedDateTime.toLocalTime().format(DateTimeFormatter.ofPattern("HH:mm:ss")));
+            utcOffsets.add(localZonedDateTime.getOffset().getId());
+        }
+
+        return new LocalDateTimeResult(localDates, localTimes, utcOffsets);
+    }
+
+
     /**
+     * Will remove this method after approval.
      * Calculates local date, local time, and UTC offset for each entry based on given UTC dates, times, latitudes, and longitudes.
      * @param map          map object (created once per upload for efficiency.
      * @param utcDates     list of UTC dates as strings (e.g., "yyyy-MM-dd")
@@ -290,7 +349,6 @@ public class TimeUtils {
      * @param longitudes   list of longitude strings
      * @return a LocalDateTimeResult containing lists of local dates, times, and UTC offsets.
      */
-
     public static LocalDateTimeResult calculateLocalDateTime(TimeZoneMap map,
                                                              ArrayList<String> utcDates, ArrayList<String> utcTimes,
                                                              ArrayList<String> latitudes, ArrayList<String> longitudes) throws Exception {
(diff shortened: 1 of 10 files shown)


src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
 
-        FlightBuilder builder = new CSVFlightBuilder(meta, doubleTimeSeries, stringTimeSeries);
+        List<String> headerLines = new ArrayList<>();
+        headerLines.add(flightInformationLine);
+        headerLines.add(secondLine);
+        headerLines.add(thirdLine);
 
-        LOG.info("Returning flight builder!");
-        return Stream.of(builder);
+        return headerLines;
```
jkarns275, line 295: This seems like a regression from them original header processing code. I would suggest you conditionally modify the unit headers for G5 data exclusively

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+     * @param headerLines
      * @return
+     * @throws IOException
+     */
+    private byte[] writeFlightToCSV(List<String[]> flightData, List<String> headerLines) throws IOException {
+        LOG.info("CSVFileProcessor - writeFlightToCSV - start");
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        OutputStreamWriter writer = new OutputStreamWriter(outputStream, StandardCharsets.UTF_8);
+        BufferedWriter bufferedWriter = new BufferedWriter(writer);
+        try {
```
jkarns275, line 327: Don't need a try block

src/main/java/org/ngafid/flights/process/FlightMeta.java
```diff
@@ -23,6 +23,33 @@ public final class FlightMeta {
...
+        this.uploaderId = other.uploaderId;
+        this.uploadId = other.uploadId;
+        this.processingStatus = other.processingStatus;
+        this.startDateTime = other.startDateTime != null ? new String(other.startDateTime) : null;
+        this.endDateTime = other.endDateTime != null ? new String(other.endDateTime) : null;
+        this.md5Hash = other.md5Hash != null ? new String(other.md5Hash) : null;
+        this.systemId = other.systemId != null ? new String(other.systemId) : null;
+        this.filename = other.filename != null ? new String(other.filename) : null;
+        this.calculated = other.calculated != null ? new String(other.calculated) : null;
+        this.suggestedTailNumber = other.suggestedTailNumber != null ? new String(other.suggestedTailNumber) : null;
```
jkarns275, line 42: Strings in java are read only, so you don't need to copy them.

src/main/java/org/ngafid/flights/process/FlightMeta.java
```diff
@@ -23,6 +23,33 @@ public final class FlightMeta {
...
+
+        // Deep copy of Airframe if not null
+        this.airframe = other.airframe != null ? new Airframes.Airframe(other.airframe.getName()) : null;
+
+        // Deep copy of AirframeType if not null
+        this.airframeType = other.airframeType != null ? new Airframes.AirframeType(other.airframeType.getName()) : null;
```
jkarns275, line 48: I think these are read only as well.

src/main/java/org/ngafid/flights/process/Pipeline.java
```diff
@@ -297,7 +297,15 @@ public synchronized void addDerivedFile(String filename, byte[] data) throws IOE
...
-        Files.createDirectories(zipFileSystemPath.getParent());
+        Path parentPath = zipFileSystemPath.getParent();
+        if (parentPath != null && !Files.exists(parentPath)) {
+            Files.createDirectories(parentPath);
+        }else{
+            String path = upload.getDerivedDirectory();
+            Path derivedPath = Paths.get(path);
+            Files.createDirectories(derivedPath);
+        }
+
```
jkarns275, line 308: I don't know what the purpose of this is.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+        return headerLines;
     }
 
     /**
-     * Gets the flight information from the first line of the file
-     * 
-     * @param reader BufferedReader for reading the first line
+     G5, G3x data has metadata formated like this: UTC Date (yyyy-mm-dd),UTC Time (hh:mm:ss),
+     This method extracts the content inside parentheses
+     If parentheses not found, returns the original input.
```
jkarns275, line 301: format

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+     * @param bufferedReader
+     * @return List of header lines (3 lines expected)
+     * @throws IOException
+     */
+    private List<String> extractHeaderLines(BufferedReader bufferedReader) throws IOException {
+        String flightInformationLine = bufferedReader.readLine();
+        String secondLine = bufferedReader.readLine();
+        String thirdLine = bufferedReader.readLine();
+
+        if(secondLine.startsWith("#")){
```
jkarns275, line 285: need spaces e.g. `if (...) {`

jkarns275: This applies to a lot of code but im not going to mark them all.

jkarns275 (changes requested):

src/main/java/org/ngafid/common/TimeUtils.java
```diff
@@ -113,14 +112,39 @@ public static String toUTC(String date, String time, String offset) {
...
 
     }
+    /**
+     * List of DateTime formats supported.
+     */
+    private static final List<DateTimeFormatter> dateTimeFormatters = Arrays.asList(
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"),
+            DateTimeFormatter.ofPattern("M/d/yyyy HH:mm:ss")
+
+    );
```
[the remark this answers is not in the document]
eightspokes: Done

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
-     * Gets the flight information from the first line of the file
-     * 
-     * @param reader BufferedReader for reading the first line
+     G5, G3x data has metadata formated like this: UTC Date (yyyy-mm-dd),UTC Time (hh:mm:ss),
+     This method extracts the content inside parentheses
+     If parentheses not found, returns the original input.
+     */
+    public static String extractContentInsideParentheses(String input) {
+
+        Pattern pattern = Pattern.compile("\\(([^)]+)\\)");
```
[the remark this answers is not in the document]
eightspokes: Done

src/main/java/org/ngafid/flights/process/FlightMeta.java
eightspokes: Done

src/main/java/org/ngafid/flights/process/FlightMeta.java
eightspokes: Airframe is a mutable object, its internal state (like not final name field) can change overtime. We probably do not expect the Airframe to change though.. Removed.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
eightspokes: Done, thanks!

src/main/java/org/ngafid/flights/process/Pipeline.java
eightspokes: It simply checks if the directory exists, and if it doesn't, it creates it. I removed else, since it is redundant.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
eightspokes: Replaced with try with resources

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+                // TODO: TimeZoneEngine initialization take significant amount of time 1-2 seconds. If we expect many G5 files,
+                //  timeZoneMap needs to be created once per program lifetime.
+                // Time zone for the US including Alaska and Hawaii
+                timeZoneMap = TimeZoneMap.forRegion(18.91, -179.15, 71.538800, -66.93457);
             }
 
-            // If we detect an invalid row before the last row of the file, or there are no valid rows.
-            if (validRows < Math.max(columns.size() - 2, 0)) {
-                throw new FatalFlightFileException(
-                        "Flight file has 0 valid rows - something serious is wrong with the format.");
```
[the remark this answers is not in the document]
eightspokes: Done

src/main/java/org/ngafid/common/TimeUtils.java
```diff
@@ -259,4 +283,93 @@ public static Date addMilliseconds(Date date, Integer milliseconds) {
...
+    /**
+     * Calculates local date, local time, and UTC offset for each entry based on given UTC dates, times, latitudes, and longitudes.
+     * @param map          map object (created once per upload for efficiency.
+     * @param utcDates     list of UTC dates as strings (e.g., "yyyy-MM-dd")
+     * @param utcTimes     list of UTC times as strings (e.g., "HH:mm:ss")
+     * @param latitudes    list of latitude strings
+     * @param longitudes   list of longitude strings
+     * @return a LocalDateTimeResult containing lists of local dates, times, and UTC offsets.
+     */
+    public static LocalDateTimeResult calculateLocalDateTime(TimeZoneMap map,
```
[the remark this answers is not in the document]
jkarns275: Not addressed

src/main/java/org/ngafid/common/TimeUtils.java
```diff
@@ -259,4 +282,171 @@ public static Date addMilliseconds(Date date, Integer milliseconds) {
...
+
+            // Try parsing using a formatter.
+            for (DateTimeFormatter formatter : DATE_FORMATTERS) {
+                try {
+                    utcDateTime = LocalDateTime.parse(dateTimeString, formatter);
+                    break;
+                } catch (DateTimeParseException e) {
+                    // Continue trying other formats
+                }
+            }
```
jkarns275, line 379: Determine the date format once and use it at every time step. Move this out of the loop.

src/main/java/org/ngafid/common/TimeUtils.java
```diff
@@ -259,4 +282,171 @@ public static Date addMilliseconds(Date date, Integer milliseconds) {
...
```
jkarns275, line 317: Determine formatter once instead of doing it once for every element inside of utcDates.

src/main/java/org/ngafid/flights/Upload.java
```diff
@@ -577,6 +577,10 @@ public String getArchiveFilename() {
         }
     }
 
+
+    public String getDerivedDirectory() {
+        return WebServer.NGAFID_ARCHIVE_DIR + "/" + fleetId + "/" + uploaderId + "/derived";
+    }
```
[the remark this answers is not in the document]
jkarns275: Still not removed.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -22,21 +29,22 @@
...
+ * @author Roman Kozulia
  */
-
 public class CSVFileProcessor extends FlightFileProcessor {
     private static final Logger LOG = Logger.getLogger(CSVFileProcessor.class.getName());
     private final List<String> headers;
     private final List<String> dataTypes;
     private final FlightMeta meta = new FlightMeta();
+    private static final Pattern PARENTHESIS_PATTERN = Pattern.compile("\\(([^)]+)\\)");
+    private static final int splitTimeInMinutes = 5; // Time difference indicates a separate flight.
```
jkarns275, line 40: `SPLIT_TIME_MINUTES`

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,408 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+                stringTimeSeries.clear();
             }
+        } catch (IOException | FatalFlightFileException | CsvException e) {
+            throw new FlightProcessingException(e);
+        } catch (SQLException e) {
+            String errorMessage = "SQL error occurred while processing the flight data for file: " + filename;
+            throw new FlightProcessingException(errorMessage, e);
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e);
+        }
```
jkarns275, line 176: Remove

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,408 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+    /**
+     *  Ands derived CSV files to the uploads folder
+     * @param splitIndices
+     * @param rows
+     * @param headerLines
+     * @param filename
+     * @throws IOException
+     * @throws SQLException
+     */
+    private void addDerivedFileToUploadFolder(List<Integer> splitIndices, List<String[]> rows, List<String> headerLines, String filename) throws IOException, SQLException {
```
jkarns275, line 191: Do you need to write this to disk? I figure it can be added directly to the zip. Doing this will also leave junk data on the filesystem if not removed properly.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,408 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
                 try {
-                    Double.parseDouble(column.get(0));
-                    doubleTimeSeries.put(name, new DoubleTimeSeries(name, dataType, column));
-                } catch (NumberFormatException e) {
-                    stringTimeSeries.put(name, new StringTimeSeries(name, dataType, column));
+                    LocalDate.parse(firstRow[0], formatter);
+                    return true; // If the date is valid, we return true
+                } catch (DateTimeParseException e) {
+                    // If the date is not valid, we return false or handle the logic accordingly
+                    return false;
```
jkarns275, line 280: Indentation seems off?

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,408 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+        return outputStream.toByteArray();
+    }
+
+    /**
+     * Splits flights based on time intervals between rows and returns flight indices.
+     * @param rows
+     * @param splitIntervalInMinutes - max time difference between rows.
+     * @return
+     */
+    public List<Integer> splitCSVIntoFlightIndices(List<String[]> rows, int splitIntervalInMinutes) {
```
jkarns275, line 405: Do split after conversion to double and string series.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,408 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+    }
+
+    /**
+     * Processes header lines from the csv flight, populates dataTypes and headers with
+     * the flight information.
+     * @param headerLines
+     * @throws FatalFlightFileException
+     */
+    private void processHeaders(List<String> headerLines) throws FatalFlightFileException {
+        if (headerLines.get(1) != null) {
```
jkarns275, line 293: This seems weird.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,408 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+                    }
+                }
+                // G5 doesn't have local date time and utcOffset column, we calculate them here
+                TimeUtils.LocalDateTimeResult localDateTimeResult = null;
+                if(isG5FlightRecorder) {
+                    StringTimeSeries utcDateSeries = new StringTimeSeries(headers.get(0), dataTypes.get(0), columns.get(0));
+                    StringTimeSeries utcTimeSeries = new StringTimeSeries(headers.get(1), dataTypes.get(1), columns.get(1));
+                    DoubleTimeSeries latitudeTimeSeries = new DoubleTimeSeries(headers.get(4), dataTypes.get(4), columns.get(4));
+                    DoubleTimeSeries longitudeTimeSeries = new DoubleTimeSeries(headers.get(5), dataTypes.get(5), columns.get(5));
+                    localDateTimeResult = TimeUtils.calculateLocalDateTimeFromTimeSeries(timeZoneMap, utcDateSeries,utcTimeSeries,latitudeTimeSeries,longitudeTimeSeries);
```
jkarns275, line 130: I would recommend not assuming the order of the columns will alwys be the same for the g5 recorder.

Also, I think you can preserve the original column names using this aliasing system:
https://github.com/travisdesell/ngafid2.0/blob/josh/refactor/src/main/java/org/ngafid/flights/process/FlightBuilder.java#L137


Basically, you will alias g5 column names to the normalized names we use (e.g. whatever g5 uses for lcl date will be aliased to Lcl Date). If this isn't sufficient, you may want to move this to its own processing step that outputs Lcl Date etc. columns and subclass the CSVFlightBuilder to add that additional processing step.

jkarns275 (changes requested):

src/main/java/org/ngafid/common/TimeUtils.java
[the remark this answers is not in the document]
eightspokes: I did address this, see alculateLocalDateTimeFromTimeSeries method. For some reason this code is not reflecting the changes. The code is marked as outdated.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +56,353 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+     * @return a list of lists of type string, where each inner list represents a separate flight
+     */
+    public List<List<String[]>> splitCSVIntoFlights(List<String[]> rows, boolean isG5flightRecorder) {
+
+        List<List<String[]>> flights = new ArrayList<>();
+        List<String[]> currentFlight = new ArrayList<>();
+        LocalDateTime lastTimestamp = null;
+
+        for (String[] row : rows) {
+            if (isG5flightRecorder && row[4].isEmpty()) {
```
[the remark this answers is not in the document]
eightspokes: This code is outdated. The new version used indices.

`  // Process each flight by splitting the series based on the indices
            for (int i = 0; i < splitIndices.size(); i++) {
                int fromIndex = splitIndices.get(i);
                int toIndex = (i == splitIndices.size() - 1) ? rows.size() : splitIndices.get(i + 1);
                List<String[]> flightRows = rows.subList(fromIndex, toIndex);
`

src/main/java/org/ngafid/flights/Upload.java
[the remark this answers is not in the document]
eightspokes: Missed it, sorry.

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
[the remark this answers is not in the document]
eightspokes: This code is outdated, please see files changed

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
eightspokes: I needed to have this method so that I have access to the header files within the scope of the parse method. 
It might also be a cleaner solution that enhances readability and maintainability.

I removed the code below so that the header files remain intact ( if that was your concern)

`  if(secondLine.startsWith("#")){
            LOG.info(logTag + " extractHeaderLines " + " # found in the beginning of the second line, removing." );
            secondLine = secondLine.substring(1);  // Remove the first character (if it's '#')
        }`

src/main/java/org/ngafid/common/TimeUtils.java
```diff
@@ -114,13 +117,30 @@ public static String toUTC(String date, String time, String offset) {
...
+        String normalizedDateTime = dateTimeString.replaceAll("\\s+", " ");  // Replaces multiple spaces with a single space
+        for (DateTimeFormatter formatter : DATE_FORMATTERS) {
+            try {
+                return LocalDateTime.parse(normalizedDateTime, formatter);
+            } catch (Exception e) {
+                // Continue trying other formats
+            }
+        }
+        // If none of the formats work, throw an exception or handle it appropriately
+        throw new IllegalArgumentException("Date format not supported: " + normalizedDateTime);
```
jkarns275, line 135: Please stop using this exception type. Here a FlightFileFormat exception would work, or you could create a new checked exception type.

src/main/java/org/ngafid/common/TimeUtils.java
```diff
@@ -114,13 +117,30 @@ public static String toUTC(String date, String time, String offset) {
...
+    }
+
     public static OffsetDateTime convertToOffset(String originalDate, String originalTime, String originalOffset,
```
jkarns275, line 138: Does this method actually get used anywhere?

src/main/java/org/ngafid/common/TimeUtils.java
jkarns275: Please use this to determine the formatter outside the body of the loop.
```
DateTimeFormatter formatter = null;
var dateTimeString = utcDates.get(0) + " " + utcTimes.get(0);
for (DateTimeFormatter candidateFormatter : DATE_FORMATTERS) {
    try {
        LocalDateTime.parse(dateTimeString, candidateFormatter);

        // Previous line didnt throw an error so this is the format we want.
        formatter = candidateFormatter;
        break;
    } catch (DateTimeParseException e) {
        // Continue trying other formats
    }
}

if (formatter == null) // None of our formats worked, should throw some sort of exception.
```

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,441 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
-                        .map(String::strip)
-                        .forEachOrdered(headers::add);
+                processFileInformation(fileInformation);
+                if (!headerLines.isEmpty()) {
+                    String[] dataTypesArray = headerLines.get(1).split(",", -1); // Second line as data types
+                    String[] headersArray = headerLines.get(2).split(",", -1);    // Third line as headers
+
+                    Arrays.stream(dataTypesArray)
+                            .map(String::strip)
+                            .map(CSVFileProcessor::extractContentInsideParentheses)
```
jkarns275, line 90: Yes

src/main/java/org/ngafid/flights/process/CSVFileProcessor.java
```diff
@@ -46,105 +54,441 @@ private CSVFileProcessor(Connection connection, byte[] bytes, String filename, P
...
+                    var dataType = dataTypes.get(j);
+
+                    try {
+                        Double.parseDouble(column.get(0));
+                        doubleTimeSeries.put(name, new DoubleTimeSeries(name, dataType, column));
+                    } catch (NumberFormatException e) {
+                        stringTimeSeries.put(name, new StringTimeSeries(name, dataType, column));
+                    }
+                }
+
```
jkarns275, line 176: So this should happen before anything is split. I think it makes the most sense to:

1. Parse all of the data, then
2. Crawl through the dates, looking for places where we may want to split.
3. Do the split.

The main parse should be in this parse method, but the splitting stuff should be in its own method. I think it will look something like this:

```
Stream<FlightBuilder> split(Map<String, DoubleTimeSeries> doubleTimeSeries, Map<String, StringTimeSeries> stringTimeSeries, ArrayList<String[]> data);
```

You may also need to pass some additional data w/ instance variables.

jkarns275 (changes requested):

jkarns275:
Should be ready to merge after these small things are addressed.

jkarns275:
I manually merged this after refactoring some things.

PR #164, opened 2024-10-14
Orange/Blue Graph Color Fix

Ilverism:
* Changed ordering of the summary page's **Percentage of Flights With Event** markers to match the **Percentage of Flights With Event Over Time** markers on the Trends page
 
 
image

image

 ---
 
 * Changed the ordering of the summary page's **Event Counts** markers to match the _Blue -> Orange -> Green -> etc._ color ordering in the other graphs, and it will also follow the ordering of the drop-down menu
 

 
image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ db/db_info.php
 db/db_info
 email_info.txt
 package-lock.json
-data/archive/*
+data/*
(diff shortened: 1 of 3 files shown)


jkarns275:
LGTM 👍

PR #165, opened 2024-10-22
Travis/raise rotorcraft transfer

travisdesell:
Python (?!?) code to do the RAISE transfer because the PHP version was broken due to issues with its SFTP library.

### DIFF
--- a/raise/transfer_rotorcraft_data.py
+++ b/raise/transfer_rotorcraft_data.py
@@ -1,0 +1,130 @@
+#!/usr/bin/python
+from datetime import datetime
+
+import MySQLdb
+import os
+import os.path
+import pysftp
+
+
+db_password = os.environ["NGAFID_DB_PASSWORD"]
+sftp_password = os.environ["RAISE_PASSWORD"]
+
+db = MySQLdb.connect(
+    host="127.0.0.1",  # your host, usually localhost
+    user="ngafid_user",  # your username
+    passwd=db_password,  # your password
+    db="ngafid",  # name of the data base
+)
+
+today = datetime.today().strftime("%Y-%m-%d")
+
+source_tail_filename = f"./system_ids_to_tails__{today}.csv"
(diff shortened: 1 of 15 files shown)


src/main/java/org/ngafid/flights/Flight.java
```diff
@@ -1993,7 +1993,8 @@ private void initialize(Connection connection, InputStream inputStream) throws F
                                 airframeName.equals("Cessna 400") ||
                                 airframeName.equals("Beechcraft A36/G36") ||
                                 airframeName.equals("Beechcraft G58") ||
-                                airframeName.equals("Beechcraft C90A King Air")) {
+                                airframeName.equals("Beechcraft C90A King Air") ||
```
jkarns275, line 1996: The rework I did to this is one of the few things I want you to take a look at.

src/main/java/org/ngafid/routes/PostMonthlyEventCounts.java
```diff
@@ -71,6 +71,8 @@ public Object handle(Request request, Response response) {
                 return gson.toJson(map.get(eventName));
             }
         } catch (SQLException e) {
+            LOG.severe(e.toString());
+            e.printStackTrace();
```
jkarns275, line 75: I figure you were trying to debug here -- I have the fix for this implemented on another branch. currently gated ofc by my reword but since its just a change to the queries we can pull it in.

Basically, I totally forgot to fix some queries and also forgot to delete event_statistics table.

raise/transfer_rotorcraft_data.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/python
```
jkarns275, line 1: I will eventually want to rewrite this in java to be honest. Liquibase will allow us to get rid of PHP, we can't allow ourselves to forswear one devil and immediately talk to another

jkarns275 (approved):
LGTM

PR #166, opened 2024-12-22
Javalin Webserver

AarC10:
Ported NGAFID from using Spark to Javalin as a webserver. Routes can now be implemented per function instead of an entire class. Some benefits include eliminating copy-paste boilerplate for each route class in Spark which included logging routes being handled, timing and mustache usage since Javalin integrates all of it. (Possibly [faster](https://javalin.io/comparisons/sparkjava#:~:text=Javalin's%20code%2Dbase%20is%20a,according%20to%20the%20TechEmpower%20Benchmarks.) too compared to Spark 👀)


jkarns275:
Does this actually work

PR #167, opened 2024-12-23
Proximity Fixes

Ilverism:
NOTE: _Separate 'temp' branch was created for this PR so that only the new changes appear. Gonna leave the original branch alone for now in case something catastrophic happens 🤷‍♂️_

**create_tables.php:**
* Flag to add update the _Rate of Closure_ table to add cascading deletes
* Aim is to prevent an error when deleting Flights with Proximity Events
* Unsure if this functionality belongs in `update_tables.php` since both use `$update_...` flags

---

**Event.java:**
* New constructor that is more suitable for the Proximity events
* Takes `startTime, endTime, startLine, endLine, severity, flightId, otherFlightId`

---

**CalculateProximity.java**

* Replaced `System` print statements with `LOG` calls
* Proximity events should now get properly attached to their respective flights; previously there were issues where, given two flights A and B with a Proximity event, the Event instances created for each respective flight would get attached to A, or one would just not get attached at all
* The lateral/vertical distances are now updated at the time when the new smallest Euclidean distance is found. Now, the Proximity Event's severity should now properly reflect `sqrt(lateral^2 + vertical^2)`

### DIFF
--- a/db/create_tables.php
+++ b/db/create_tables.php
@@ -12,6 +12,7 @@ $update_turn_to_final = false;
 $update_visited_airports = false;
 $update_uploads_for_raise = false;
 $update_rate_of_closure = false;
+$update_rate_of_closure_cascade_delete = true;
 $create_airsync = false;
 $create_event_metadata = false;
 //need to drop and reload these tables for 2020_05_16 changes
@@ -689,6 +690,9 @@ if (!$update_turn_to_final) {
 }
 
 if (!$update_rate_of_closure) {
+
+    echo "Creating 'rate_of_closure' table...\n";
+
     $query = "CREATE TABLE `rate_of_closure` (
         `id` INT(11) NOT NULL AUTO_INCREMENT,
         `event_id` INT(11) NOT NULL,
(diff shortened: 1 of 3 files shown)


db/create_tables.php
```diff
@@ -689,19 +690,53 @@
...
+
     $query = "CREATE TABLE `rate_of_closure` (
         `id` INT(11) NOT NULL AUTO_INCREMENT,
         `event_id` INT(11) NOT NULL,
         `size` INT(11) NOT NULL,
         `data` MEDIUMBLOB,
         
         PRIMARY KEY(`id`),
-        FOREIGN KEY(`event_id`) REFERENCES events(`id`)
+        FOREIGN KEY(`event_id`) REFERENCES events(`id`) ON DELETE CASCADE   -- Source Event deleted => Delete rate of closure
```
AarC10, line 703: You might want to update this in Josh's liquibase PR as well

src/main/java/org/ngafid/proximity/CalculateProximity.java
```diff
@@ -5,104 +5,147 @@
...
+import java.util.HashMap;
+import java.util.logging.Logger;
+import java.util.Map;
+
 import org.ngafid.events.EventMetaData;
 import org.ngafid.events.EventStatistics;
 import org.ngafid.airports.Airports;
 
+//import com.mysql.cj.log.Log;
+
```
AarC10, line 24: Can this line just be removed?

src/main/java/org/ngafid/proximity/CalculateProximity.java
```diff
@@ -5,104 +5,147 @@
...
         return distanceFt;
+        
     }
 
-    public static double calculateLateralDistance(double flightLatitude, double flightLongitude, double otherFlightLatitude,
-                                           double otherFlightLongitude) {
+    public static double calculateLateralDistance(double flightLatitude, double flightLongitude,    double otherFlightLatitude, double otherFlightLongitude) {
        
-        double lateralDistance = Airports.calculateDistanceInFeet(flightLatitude, flightLongitude, otherFlightLatitude, otherFlightLongitude);
+        double lateralDistance = Airports.calculateDistanceInFeet(flightLatitude, flightLongitude,      otherFlightLatitude, otherFlightLongitude);
```
AarC10, line 76: Weird stuff going on with the spacing here

src/main/java/org/ngafid/proximity/CalculateProximity.java
```diff
@@ -541,39 +727,49 @@ public static void main(String[] arguments) {
...
                 double elapsed_seconds = (double)Duration.between(start, end).toMillis() / 1000.0;
                 double average_seconds = ((double) elapsed_seconds) / (double)count;
-                System.out.println("calculated " + count + " adjacency evaluations in " + elapsed_seconds + " seconds, averaged: " + average_seconds + " seconds per flight");
-                System.out.println("avg time matched flights: " + ((double)timeMatchFlights / (double) count) + ", avg loc matched flights: " + ((double)locMatchFlights / (double)count));
-                System.out.println("evnets found:"  + eventsFound);
-                //System.exit(1);
+
+                LOG.info("calculated " + count + " adjacency evaluations in " + elapsed_seconds + " seconds, averaged: " + average_seconds + " seconds per flight");
+                LOG.info("avg time matched flights: " + ((double)timeMatchFlights / (double) count) + ", avg loc matched flights: " + ((double)locMatchFlights / (double)count));
+                LOG.info("evnets found:"  + eventsFound);
```
AarC10, line 746: Typo

AarC10 (changes requested):

db/create_tables.php
jkarns275: True. Just keep it in mind and make a PR to that branch, you don't have to do anything to this branch.

PR #169, opened 2025-01-13
Merge Proximity Fixes → Liquibase

Ilverism:
**Copy of  [Proximity Fixes #167](https://github.com/travisdesell/ngafid2.0/pull/167) (Main ← Proximity Fixes)**

--- 

NOTE: _Separate 'temp' branch was created for this PR so that only the new changes appear. Gonna leave the original branch alone for now in case something catastrophic happens 🤷‍♂️_

**create_tables.php:**
* Flag to add update the _Rate of Closure_ table to add cascading deletes
* Aim is to prevent an error when deleting Flights with Proximity Events
* Unsure if this functionality belongs in `update_tables.php` since both use `$update_...` flags

---

**Event.java:**
* New constructor that is more suitable for the Proximity events
* Takes `startTime, endTime, startLine, endLine, severity, flightId, otherFlightId`

---

**CalculateProximity.java**

* Replaced `System` print statements with `LOG` calls
* Proximity events should now get properly attached to their respective flights; previously there were issues where, given two flights A and B with a Proximity event, the Event instances created for each respective flight would get attached to A, or one would just not get attached at all
* The lateral/vertical distances are now updated at the time when the new smallest Euclidean distance is found. Now, the Proximity Event's severity should now properly reflect `sqrt(lateral^2 + vertical^2)`


jkarns275:
I manually merged this.

PR #172, opened 2025-01-16
Get User Email Preferences Fixes and Uses Form Params

AarC10:
Fixes issues causing the preferences page not to load most likely due to merge conflict issues and the email preferences not loading. Also uses form parameters instead of query parameters as requested by @jkarns275


jkarns275:
Based + Lint pilled.

PR #174, opened 2025-01-25
Roman/vfr chart processor

eightspokes:
**Objective:**  
This pull request introduces the VFR chart processing feature that downloads TIF files from the FAA website and generates raster tiles. If the `charts` folder or any of its subfolders (`sectional`, `terminal-area`, `ifr-enroute-low`, `ifr-enroute-high`) are missing, meaning the charts are not present, the script will automatically download the required charts using the closest release date, process them into raster tiles, and start the chart server once the downloads and processing are complete.  

**Testing:**  
The script automatically starts when the WebServer is launched.
When the program is started for the first time and the chart folder is not present, the script will trigger the download and processing of TIF files. The process may take approximately **1 hour** to complete. You can monitor the progress in the log files:  
- `chart_processor.log`: Logs related to file downloading and processing.  
- `chart_server.log`: Logs related to the server startup.  

To force-start the process of downloading and processing charts, simply delete charts folder and restart the WebServer.

The server will run at midnight daily and check if today is the update date. If it is, the script will start the download and update process. The update schedule is stored in **services/chart_processor/chart_service_config.json**

**Dependencies:**  
You will need to install **gdalwarp command-line tool** used for geospatial transformation. Unfortunately, gdalwarp cn not be installed using pip. 

To install **gdalwarp** run:
On macOS: `brew install gdal`
On Linux: `sudo apt-get install gdal-bin`
On Windows: Download from `https://www.gisinternals.com/ `or install via OSGeo4W

The system will install other python dependency using **requirement.txt file**

**Configuration:**  
Currently, host/port for chart service are configured  for `"localhost: 8187 "`. 
Host/port configuration can be changed in: **chart_service_config.json** You will also need to **change host/port in src/main/javascript/map.js**
E.g:  ` http://localhost:8187/terminal-area/{z}/{x}/{-y}.png`


jkarns275:
Just glanced at this -- do all of the shape files need to be included? Are they downloaded automatically?

I just ask because 700 files is kind of a lot.

eightspokes:
> Just glanced at this -- do all of the shape files need to be included? Are they downloaded automatically?
> 
> I just ask because 700 files is kind of a lot.


@jkarns275  We need all  of those files for the shapefiles to work correctly. A shapefile is not just a single file but a collection of files that work together to store spatial and attribute data. 
So for each polygon we have a separate folder with 5 files: .shp (Shapefile Geometry) , .shx (Shape Index File), .dbf (Attribute Table), .prj (Projection File), .cpg (Character Encoding File).

We have about 150 polygons so 700 files sounds about right.

But the good thing is that the files are very small, all shape files take about 3 Mb of memory.

PR #175, opened 2025-01-27
Updated/Distinct HTML Titles

Ilverism:
Quick change to display distinct browser tab titles per page.

Changed version matches the common format of _Page Name_ - _Site Name_.

Should make navigation and testing easier, especially when multiple site tabs are open at once.

--- 

Changed: 
image

Original:
image


jkarns275:
Nice

jkarns275:
@AarC10 looks like compile and build CI is broken because of JUnit or something?

AarC10:
Will take a look when Im  out of class

PR #179, opened 2025-02-07
Fix Failing CI From Running Maven Tests

AarC10:
CI started to mysteriously break. Hoping this fixes it. Also somewhat alphabetized and updated JUnit versions


jkarns275:
pog

PR #180, opened 2025-02-10
Use fa-spinner instead of ajax-loader.gif

Ilverism:
* Centered loading icon
* Dark mode support for transparent layer behind icon
* Replaced `ajax-loader.gif` spinner with `fa-spinner`

**Before**

image

**After**

image


jkarns275:
powerful

PR #181, opened 2025-02-14
Trends Page Fixes (+EventDefinitions fix)

Ilverism:
Fixes for stuff on the Trends page that broke (presumably because of Plotly updates)

* Manual formatting for fetched severity types in EventDefinition constructor to match enum names (i.e. capitalization and inserting underscores)
* Re-added missing plot titles to both trends plots
* Fixed bug causing the trends plots to have heights of 1,000,000+ pixels

_(Similar issues probably on Severities page, will check that too)_


jkarns275 (approved):
lgtm

PR #186, opened 2025-02-27
Chart service working on the most recent source code.

eightspokes:
**Objective:**  
This pull request introduces the VFR chart processing feature that downloads TIF files from the FAA website and generates raster tiles. If the `charts` folder or any of its subfolders (`sectional`, `terminal-area`, `ifr-enroute-low`, `ifr-enroute-high`) are missing, meaning the charts are not present, the script will automatically download the required charts using the closest release date, process them into raster tiles, and start the chart server once the downloads and processing are complete.  

**Testing:**  
The script automatically starts when the WebServer is launched.
When the program is started for the first time and the chart folder is not present, the script will trigger the download and processing of TIF files. The process may take approximately **1 hour** to complete. You can monitor the progress in the log files:  
- `chart_processor.log`: Logs related to file downloading and processing.  
- `chart_server.log`: Logs related to the server startup.  

To force-start the process of downloading and processing charts, simply delete charts folder and restart the WebServer.

The server will run at midnight daily and check if today is the update date. If it is, the script will start the download and update process. The update schedule is stored in **services/chart_processor/chart_service_config.json**

**Dependencies:**  
You will need to install **gdalwarp command-line tool** used for geospatial transformation. Unfortunately, gdalwarp cn not be installed using pip. 

To install **gdalwarp** run:
On macOS: `brew install gdal`
On Linux: `sudo apt-get install gdal-bin`
On Windows: Download from `https://www.gisinternals.com/ `or install via OSGeo4W

The system will install other python dependency using **requirement.txt file**

**Configuration:**  
Currently, host/port for chart service are configured  for `"localhost: 8187 "`. 
Host/port configuration can be changed in: **chart_service_config.json** You will also need to **change host/port in src/main/javascript/map.js**
E.g:  ` http://localhost:8187/terminal-area/{z}/{x}/{-y}.png`


eightspokes:
@jkarns275 Since the file structure has changed quite a bit, I created a integrated the chart service into the most recent code. All seems to be working fine. Please give it a run when you have a moment. Once dependencies are installed and you see that tif files are downloading, I would give leave it running overnight. I changed zoom level to increase chart quality, so the tile pyramid may take some time to process.

Once charts are built, the second run will not require any processing and will be done without delay.

jkarns275:
@eightspokes I think this is good however a README.md addendum w/ some brief deployment instructions would be good. You might just want to create one in the `services/chart_processor` directory and point to it from the base `README.md`.

jkarns275:
Conflicts

PR #187, opened 2025-02-28
Persistent Login After Restarting Webserver

AarC10:
By default, the webserver will cache sessions so when the server restarts, there should be no need to log back in, unless the user intentionally logs out. This can be opted out of by adding `export DISABLE_PERSISTENT_SESSIONS=true` as an environment variable


jkarns275:
You snapped on this one

AarC10:
image

PR #188, opened 2025-03-02
Define views and materialized views for fast event and flight statistics.

jkarns275:
# Summary
The way we compute event statistics is too slow, and it is impossible to run some of these queries in a reasonable time frame. To avoid this, I have implemented what is often called a "materialized view." It is like a normal view, which is analogous to a macro for a `SELECT` query, except the results are cached.

I define materialized views for monthly event counts, flight counts, and flight time. Normal views are defined on top of these materialized views, which should run reasonably fast.

This should avoid the weirdness with the old event statistics table, while also being really fast -- at most, MySQL will only have to aggregate several dozen rows rather than going over the entire events or flights table.

# Deployment
These materialized views take a while to compute, so they will only be computed occasionally. This means some of the information will be out of date if an upload is deleted or processed, for example. They will be computed hourly or daily, which can be done with a cronjob and a liquibase command:

`mvn liquibase:update -Dliquibase.contexts="daily-materialized-views"`
`mvn liquibase:update -Dliquibase.contexts="hourly-materialized-views"`

If you don't run these after processing data, the event statistics will be out of date.


src/main/java/org/ngafid/uploads/process/format/CSVFileProcessor.java
```diff
@@ -310,7 +310,7 @@ void processMetaData(BufferedReader reader) throws FatalFlightFileException {
         // This is where we can integrate airframe name input from user.
         // Check if flight information contains airframe_name and system_id, if not, put dummy values (for testing).
         if (!values.containsKey("airframe_name")) {
-            values.put("airframe_name", "Cessna 172S");
+            values.put("airframe_name", "Unknown");
             LOG.severe("!!! TESTING ONLY: Log: airframe_name is missing, setting to DummyAirframe - Cessna 172S.");
```
AarC10, line 314: LOG needs to be updated to align with the above change

AarC10 (changes requested):

src/main/java/org/ngafid/uploads/process/format/CSVFileProcessor.java
jkarns275: I wonder if we should even allow G5 data to be imported like this

PR #189, opened 2025-03-06
Interface Updates for Cesium Fixes (+Other adjustments)

Ilverism:
* Webpack changes to fix problems with builds taking like 60 seconds whenever something changes _(think this had to do with Babel recompiling already bundled code)_
* Removed the 2 replay buttons on the Paginator and disabled the replay camera button 🎦 _(assuming we aren't going to need those since we can view the Cesium replays on the same page anyways?)_
* Fixed a problem with the error modal not being displayed by forcing the title and message to be Strings _(seemed to mostly be a problem with the empty/malformed JSON data in ajax calls)_
* Made changes to the list of displayed filters _(NOTE: still having some small issues with this, will revisit)_
* Added a dropdown menu to the Cesium map to change the resolution scale
* Fixed Flight tracking button for Cesium
* Added a button next the tracking button to automatically move the Cesium viewer timeline playhead to the start of the Flight
* Added a button/indicator for the tracked Flight to the Cesium map, will stop tracking the currently tracked Flight when clicked
* Other misc UI changes for the Flights page

Will probably still require another pass to clean up some of the old/deprecated stuff from the old Cesium version. Should kill the www_php folder while we're at it too.

---

image

---

image

---

image

---

image

---

image

---

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,5 @@ db/db_info
 email_info.txt
 package-lock.json
 data/*
-src/main/resources/liquidbase.properties
+src/main/resources/liquibase.properties
+src/main/resources/public/dist/*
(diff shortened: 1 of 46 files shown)


jkarns275:
Looks good. There are a few more things that @e911 needs to do to fix the branch before we can merge to main.

jkarns275:
Merge conflct

PR #192, opened 2025-03-12
LRU Terrain Cache

AarC10:
Cache terrain maps and remove the least recently used terrains from the cache based on access order.

### DIFF
--- a/src/main/java/org/ngafid/common/terrain/TerrainCache.java
+++ b/src/main/java/org/ngafid/common/terrain/TerrainCache.java
@@ -9,7 +9,7 @@ import java.util.logging.Logger;
 
 public final class TerrainCache {
     public static final String TERRAIN_DIRECTORY;
-    private static final int MAX_CACHE_SIZE;
+    private static int MAX_CACHE_SIZE;
     private static final Logger LOG = Logger.getLogger(TerrainCache.class.getName());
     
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/ngafid/common/terrain/TerrainCache.java
```diff
@@ -1,31 +1,52 @@
...
 public final class TerrainCache {
-
-    private static final SRTMTile[][] TILES = new SRTMTile[180][360];
     public static final String TERRAIN_DIRECTORY;
+    private static int MAX_CACHE_SIZE;
+    private static final Logger LOG = Logger.getLogger(TerrainCache.class.getName());
+    
+
+    // Load factor is arbitrary because we always remove the eldest entry
+    private static final Map<Coordinate, SRTMTile> TILE_CACHE = new ConcurrentHashMap<>(new LinkedHashMap<>(MAX_CACHE_SIZE, 0.75f, true) {
```
jkarns275, line 17: This constructor to ConcurrentHashMap creates a copy of the mappings in the supplied map -- I don't think this will work.

It looks like there is no standard `LinkedConcurrentHashMap` and using a lock to access the map each and everytime is WEAK. You will have to implement the cache-order thing yourself I fear -- enjoy.

jkarns275 (changes requested):

src/main/java/org/ngafid/common/terrain/TerrainCache.java
```diff
@@ -13,14 +12,14 @@ public final class TerrainCache {
...
-            return size() > MAX_CACHE_SIZE;
-        }
-    });
-    
+     private static final ThreadLocal<LinkedHashMap<Coordinate, SRTMTile>> THREAD_LOCAL_TILE_CACHE =
+        // Load factor is arbitrary because we always remove the eldest entry
+        ThreadLocal.withInitial(() -> new LinkedHashMap<>(MAX_CACHE_SIZE, 0.75f, true) {
+            @Override
+            protected boolean removeEldestEntry(Map.Entry<Coordinate, SRTMTile> eldest) {
+                return size() > MAX_CACHE_SIZE;
```
jkarns275, line 20:
```
if (MAX_CACHE_SIZE > 0)
    return size() > MAX_CACHE_SIZE;
else
    return false;
````

To allow for an unbounded cache (useful in prod).

AarC10: You got it chief

PR #196, opened 2025-04-08
Logging api request info in db

e911:


jkarns275:
This pretty much looks good, just tweak some of the datatypes in table. I think the status should be int and the ip address can be `VARBINARY(16)` (since we can't assume IPv4 or IPv6).

PR #198, opened 2025-04-23
ESLint + Command

Ilverism:
Added ESLint config and a command to use it to check for warnings & errors:
`npm run check`

--- 
Example Output:
image


jkarns275:
@Ilverism Fix merge conflicts

PR #200, opened 2025-04-28
Generic event definition tables + spacing for event definition name groups

Ilverism:
- getAllEventDescriptions will now also retrieve all event definitions with an airframe ID of 0
- Two tables are used on the Event Definitions page for generic & airframe-specific events
- Spacing is now applied to group different event names in the specific event defs table

image


jkarns275:
🥀

PR #206, opened 2025-05-12
Parquet Processor Integration

eightspokes:


eightspokes:
@jkarns275 If you think this is okay? maybe I can merge this pull request?

jkarns275:
Conflicts with your other PR. Should be an easy fix

ngafid-data-processor/src/main/java/org/ngafid/processor/format/ParquetFlightBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+                parquetSteps.stream()
+                        .map(factory -> factory.create(connection, this))
+                        .filter(step ->
+                                step.getOutputColumns().stream()
+                                        .noneMatch(x ->
+                                                getDoubleTimeSeriesMap().containsKey(x) ||
+                                                        getStringTimeSeriesMap().containsKey(x)
+                                        )
+                        )
+                        .collect(Collectors.toCollection(ArrayList::new));
```
jkarns275, line 41: I don't think this is necessary anymore? And if it is, it would only apply to AltAGL in which case I recommend you just add an if statement.

PR #209, opened 2025-05-15
Addressed proximity bugs, needs testing

eightspokes:
This pull request addressed issues described in a message from Travis:

1. When doing the querying by the min/max lat/lon bounds for other potential proximity flights, make sure to increase the max and decrease the min by the proximity maximum value (e.g., 1000ft) - this captures edge cases where the min/max boxes for flights don't overlap but they still come within however many feet of each other

2. Update the required columns to not only be for fixed wing aircraft.  I think we can just require AltAGL > 50 ft (we should discuss and bring up on the thursday meeting) -- we need to make sure we can capture prox events between fixed wing and UAS or rotorcraft still show up

3 Make sure whatever was causing flights to have prox events with themselves is fixed

### DIFF
--- a/ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/FlightTimeLocation.java
+++ b/ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/FlightTimeLocation.java
@@ -5,7 +5,6 @@ import org.ngafid.core.flights.Flight;
 import org.ngafid.core.flights.Parameters;
 import org.ngafid.core.flights.StringTimeSeries;
 import org.ngafid.core.util.filters.Pair;
-import org.ngafid.processor.format.CSVFileProcessor;
 
 import java.io.IOException;
 import java.sql.Connection;
@@ -173,9 +172,6 @@ public final class FlightTimeLocation {
         LOG.info("Buffered overlap check: " + overlap);
         return overlap;
     }
-
-
-
 
 
     public boolean isValid() {
--- a/ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/FlightTimeLocation.java
+++ b/ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/FlightTimeLocation.java
@@ -98,6 +98,16 @@ public final class FlightTimeLocation {
         valid = true;
     }
 
+    // Constructor for testing only. Omits database connection
+    public FlightTimeLocation(double minLat, double maxLat, double minLon, double maxLon) {
+        this.minLatitude = minLat;
+        this.maxLatitude = maxLat;
+        this.minLongitude = minLon;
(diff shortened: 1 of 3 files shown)


ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/FlightTimeLocation.java
```diff
@@ -63,33 +67,13 @@ public FlightTimeLocation(Connection connection, Flight flight) throws SQLExcept
             return;
         }
 
-        // then check and see if this was actually a flight (RPM > 800)
-        Pair<Double, Double> minMaxRPM1 = DoubleTimeSeries.getMinMax(connection, flightId, "E1 RPM");
```
eightspokes, line 67: Travis suggested to only require altitude for validation.

From Travis:

update the required columns to not only be for fixed wing aircraft.  I think we can just require AltAGL > 50 ft (we should discuss and bring up on the thursday meeting) -- we need to make sure we can capture prox events between fixed wing and UAS or rotorcraft still show up

ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/FlightTimeLocation.java
```diff
@@ -153,10 +136,44 @@ public boolean getSeriesData(Connection connection) throws SQLException {
     }
 
     public boolean hasRegionOverlap(FlightTimeLocation other) {
-        return other.maxLatitude >= this.minLatitude && other.minLatitude <= this.maxLatitude
-                && other.maxLongitude >= this.minLongitude && other.minLongitude <= this.maxLongitude;
+
```
eightspokes, line 139: From Travis:

when doing the querying by the min/max lat/lon bounds for other potential proximity flights, make sure to increase the max and decrease the min by the proximity maximum value (e.g., 1000ft) - this captures edge cases where the min/max boxes for flights dont overlap but they still come within however many feet of each othere

ngafid-data-processor/src/main/java/org/ngafid/processor/events/proximity/ProximityEventScanner.java
```diff
@@ -248,6 +256,7 @@ public List<Event> processFlight(Connection connection, Flight flight) throws SQ
 
         List<Event> allEvents = new ArrayList<>();
         for (Flight otherFlight : potentialFlights) {
+            if (otherFlight.getId() == flight.getId()) continue;
```
jkarns275: insane that this is the source of the self-proximity event bug

eightspokes: I hope this was the cause..

eightspokes:
@jkarns275 If you think this is okay? maybe I can merge this pull request?

PR #210, opened 2025-05-19
Bug Report Page

Ilverism:
image
image

---
* Added a new bug report page
* ⚠ Requires a new Email Type, so we need to run `generate_email_types` again
* Accessible in the drop-down menu under the Account section in the navbar
* Report consists of a title and description
* The user will be BCC'd if the checkbox is checked (enabled by default)
* The report will be emailed to `NGAFID_ADMIN_EMAILS`
* Left some commented-out code for the GitHub issue functionality if we ever want that
---
* Updated the Confirm Modal to match the previous changes to the Error Modal
* The 'confirm' button in the Confirm Modal will now be hidden if there is no confirmation method supplied
---
* Added class for HTTP codes
---
* Added typing stuff to support injected `user` value in TypeScript file
---
* Added kafka quickstart files & webpack cache to gitignore


jkarns275:
I think we should CC the user, and not give an option to send anonymously lol

Ilverism:
> I think we should CC the user, and not give an option to send anonymously lol

fixd

PR #211, opened 2025-05-23
heatmap setup

eightspokes:
This PR is introduces proximity event heat map.

Currently the algorithm works as follows:

1. Get all the proximity events from database. 
2. Get the flight ids and the time period of proximity events.
3. Retrieve coordinates from DoubleTimeSeries for correspondent flight / time period.
4. Display  points of each proximity event on the heat map.

Screenshot 2025-06-05 at 5 53 09 PM

When we zoom in, we can see individual points. They are clickable:

Screenshot 2025-06-05 at 5 54 08 PM

Screenshot 2025-06-05 at 5 39 37 PM

I would appreciate you feedback on the functionality and design.


eightspokes:
created new PR for this

PR #212, opened 2025-05-23
Restored parquet processor, changed pom

eightspokes:


eightspokes:
@jkarns275 I hope the dependency won't cause any problems this time. I've also added additional guards for preventing insertion incorrect proximity events (flight with itself). 

If you have not launched new VM, maybe we could merge this and see if the problem is gone?

PR #235, opened 2025-09-09
Roman/feature/2fa implementation

eightspokes:
2 Factor Authorization


ngafid-core/src/main/java/org/ngafid/core/Config.java
```diff
@@ -62,3 +62,5 @@ public static String getEnvironmentVariable(String key) {
         return value;
     }
 }
+// Trigger workflow
+// Test trigger
```
AarC10, line 66: ?

ngafid-core/src/main/java/org/ngafid/core/accounts/User.java
```diff
@@ -89,6 +95,28 @@ private User(ResultSet resultSet) throws SQLException {
...
+        try {
+            backupCodes = resultSet.getString(16);
+        } catch (SQLException e) {
+            backupCodes = null;
+        }
+        try {
+            twoFactorSetupComplete = resultSet.getBoolean(17);
+        } catch (SQLException e) {
+            twoFactorSetupComplete = false;
+        }
```
AarC10, line 119: Cant you just set the 2FA fields to the failure fields above the try-catch and have a single try-catch block?

ngafid-www/src/main/kotlin/org/ngafid/www/services/TwoFactorAuthService.kt
```diff
@@ -0,0 +1,184 @@
...
+     * @param code The plaintext backup code to verify
+     * @param hashedCodes List of hashed backup codes from the database
+     * @return true if the code is valid, false otherwise
+     * @see #hashBackupCode(String)
+     * @see #generateBackupCodes()
+     * @throws IllegalArgumentException if code is null or hashedCodes is null
+     */
+    fun verifyBackupCode(code: String, hashedCodes: List<String>): Boolean {
+        val hashedCode = hashBackupCode(code)
+        return hashedCodes.contains(hashedCode)
```
AarC10, line 182: Can just chain the call here

pom.xml
```diff
@@ -118,7 +118,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-kotlin</artifactId>
-            <version>2.19.0</version>
+            <version>2.17.0</version>
```
AarC10, line 121: Whats the reason for downgrading software versions?

AarC10 (approved):

PR #237, opened 2025-09-10
Multiple Fleets Per User

Ilverism:
* Allows existing users to accept invites from managers of other fleets.
* Users can then accept invites from other fleets

---

_Accepting / Declining fleet invites in the User Preferences page:_

image

* Incoming invites from other users will appear as badges in a Fleet Invites table
* Users can Accept / Decline individual invites
* A confirm modal will appear, prompting the user to confirm the Accept / Decline actions
* After accepting an invite, the user will be given View access to that fleet

---

_Selecting / Leaving fleets in the User Preferences page:_

image

* Selectable fleets are displayed as badges with the fleet's name and access level displayed
* The user's current fleet is highlighted in blue
* The **Leave Current** button allows the user to permanently leave the currently selected fleet (unless reinvited)
  * After leaving a fleet, users will automatically be swapped to another fleet they have access to (prioritizing one with the highest access level)
  * Users cannot leave a fleet they have Manager level access in
  * Users cannot leave a fleet if there wouldn't be any managers left afterwards (redundancy)
  * Users cannot leave a fleet if there are no other fleets they have Manager/Upload/View access to
* If a user loses access to a fleet (i.e., Waiting/Denied), it will be grayed out and no longer be selectable

image

---

_Change Fleet from Waiting page:_

image

* In the case a user ends up logged in with a selected fleet they don't have access to, they can now select fleets from a dropdown menu inside the Waiting page
* Fleets with Waiting/Denied access levels are not selectable
* After selecting a valid fleet in the dropdown menu, they will automatically be logged in with that fleet selected

---

_Quick Select Fleet from Navbar:_

image

* Users can quickly change their fleet from any page with the Signed-In Navbar
* Fleets with the Waiting/Denied access types will not appear in the dropdown menu at all
* Selecting a fleet will trigger a page reload

### DIFF
--- a/ngafid-core/src/main/java/org/ngafid/core/accounts/User.java
+++ b/ngafid-core/src/main/java/org/ngafid/core/accounts/User.java
@@ -11,14 +11,13 @@ import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-
 public final class User implements Serializable {
+
     private static final Logger LOG = Logger.getLogger(User.class.getName());
 
     /**
-     * The following variables hold all the user information that is stored in the
-     * 'user'
-     * table in the database.
+     * The following variables hold all the user information that is stored in
+     * the 'user' table in the database.
      */
     private int id = -1;
     private String email;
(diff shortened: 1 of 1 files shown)


eightspokes (approved):
Minor improvements suggested

PR #248, opened 2025-10-24
Roman test flights

eightspokes:


ngafid-core/src/main/java/org/ngafid/core/flights/Airframes.java
```diff
@@ -254,7 +254,9 @@ public static void setAirframeFleet(Connection connection, int airframeId, int f
         if (fleetAirframes.contains(key))
             return;
         else {
-            String queryString = "REPLACE INTO fleet_airframes SET fleet_id = ?, airframe_id = ?";
```
eightspokes: Bug. Invalid query

ngafid-core/src/main/java/org/ngafid/core/flights/Flight.java
```diff
@@ -530,15 +530,15 @@ private static Set<Integer> getTagIds(Connection connection, int flightId) throw
      * @return a String that is usable in a SQL query
      */
     private static String idLimStr(Set<Integer> ids, boolean complement) {
-        StringBuilder sb = new StringBuilder("WHERE ID " + (complement ? "!" : "") + "= ");
```
eightspokes: Flight table has column id, not ID

Column: id

PR #278, opened 2026-03-03
Fix Flights page: map not showing, empty error modal

eightspokes:
# Problem
- Map and flights did not show on the Flights page (/protected/flights).
- Console showed "Map instance already exists, initialization aborted" and an error modal with empty title/message.

## Cause
- Map was initialized twice (DOMContentLoaded and FlightsPage componentDidMount). The first init ran while the map container was hidden, so the map had zero size and the correct init was skipped.
- Error modal could be shown with empty title/message from the API.

## Changes
- **map.js:** Guard against double `initializeMap()`; require `#map` in DOM; remove DOMContentLoaded auto-init so only the page's componentDidMount initializes the map.
- **error_modal.js:** Only open the modal when title or message is non-empty; remove noisy render log; safe handling of null/undefined.
- **flights.js:** Use fallbacks when showing API errors (`response.errorTitle || "Error"`, `response.errorMessage || ""`).

### DIFF
--- a/ngafid-core/src/main/java/org/ngafid/core/heatmap/HeatmapPointsProcessor.java
+++ b/ngafid-core/src/main/java/org/ngafid/core/heatmap/HeatmapPointsProcessor.java
@@ -395,6 +395,102 @@ public class HeatmapPointsProcessor {
         return eventMap;
     }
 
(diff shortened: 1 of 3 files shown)


eightspokes:
created by mistake(wrong branch)

PR #302, opened 2026-03-26
Fix login submit button issue

e911:


Ilverism:
lgtm

Confirmed working on 4 diff. browsers (though I was never able to recreate the issue Brandon reported in the first place)

PR #305, opened 2026-04-10
Allow creation of new Fleets without needing a new account

Ilverism:
* pom.xml: Bumped Kotlin version from 2.3.10 to 2.3.20

---

* Preferences Page: Users can now create new fleets with a given name
  * Will automatically switch the user to this fleet after creation
  * Users will be a manager of this fleet
* Fixed an issue with new accounts where their selected fleet was not being properly set

image
image


Andrewlee23 (approved):
lgtm

PR #326, opened 2026-06-11
Darren/upload api endpoint

dcr8024:
**Successful Upload**
image

**Invalid Token**
image

**No Access To fleet**
image


ngafid-core/src/main/java/org/ngafid/core/flights/Airframes.java
eightspokes, line 1: Please drop CSVFileProcessor.java and Airframes.java from this PR. These tails are in tail_airframe_registry, so uploads already go through RotorcraftCSVFileProcessor — both changes are redundant here.

ngafid-core/src/main/java/org/ngafid/core/flights/Airframes.java
```diff
@@ -73,6 +73,8 @@ private Airframes() {}
 
     public static final String AIRFRAME_QUEST_KODIAK_100 = "Quest Kodiak 100";
 
+    public static final String AIRFRAME_AW119 = "AW-119";
```
eightspokes: Please drop CSVFileProcessor.java and Airframes.java from this PR. These tails are in tail_airframe_registry, so uploads already go through RotorcraftCSVFileProcessor — both changes are redundant here.

ngafid-data-processor/src/main/java/org/ngafid/processor/format/CSVFileProcessor.java
eightspokes, line 1: Please drop CSVFileProcessor.java from this PR. This is redundant here.

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,314 @@
...
+                            existing.getFilename(),
+                            "You have already uploaded a file with identical contents."));
+                    return;
+                }
+
+                Upload upload = Upload.createNewUpload(
+                        connection,
+                        uploaderId,
+                        fleetId,
+                        file.filename(),
```
eightspokes: Please sanitize the upload filename before createNewUpload(), same as the web upload API. Right now file.filename() is passed through raw; names with spaces, parentheses, or other special characters are rejected (or normalized) on /api/upload via UploadRoutes.sanitizeUploadFilename() but not here. Sanitize, then reject with 400 if the name still doesn’t match ^[a-zA-Z0-9_.-]*$.

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,314 @@
...
+
+        Path tempFile = null;
+        try {
+            tempFile = Files.createTempFile("ngafid-api-upload-", ".tmp");
+            String md5Hash = streamToTempAndHash(file, tempFile);
+
+            try (Connection connection = Database.getConnection()) {
+                Upload existing = Upload.getUploadByUser(connection, uploaderId, md5Hash);
+                if (existing != null) {
+                    ctx.status(409).json(new DuplicateUploadResponse(
```
eightspokes: Duplicate handling is stricter than /api/upload. The web API only treats an upload as a duplicate when status is UPLOADED or already processed; otherwise it returns the existing row so the client can retry. This route always returns 409 if (uploader_id, md5_hash) exists, so API clients can’t re-send the same zip after UPLOADING_FAILED or a failed import without manually deleting the upload first. 

Please align with UploadRoutes.postNewUpload — 409 only for finished/processed uploads; allow retry for everything else.

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,314 @@
...
+
+                Upload upload = Upload.createNewUpload(
+                        connection,
+                        uploaderId,
+                        fleetId,
+                        file.filename(),
+                        "api-" + UUID.randomUUID(),
+                        Upload.Kind.FILE,
+                        file.size(),
+                        1,
```
eightspokes, line 199: numberChunks = 1. That doesn’t match how the web UI handles large files (2 MB chunks via /api/upload + chunk PUTs). For very large zips we should add bearer-authenticated chunked upload (init + chunk endpoints) and reuse ProcessUpload.tryCombinePieces()

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,314 @@
...
+                    upload.getMd5Hash(), upload.getStartTime(), upload.getEndTime(),
+                    upload.getFleetId()));
+        } catch (SQLException e) {
+            ctx.status(500).json(new ErrorResponse(e));
+        }
+    }
+
+    private static String streamToTempAndHash(UploadedFile file, Path dest) throws IOException {
+        MessageDigest md;
+        try { md = MessageDigest.getInstance("MD5"); }
```
eightspokes: reimplements MD5 hex encoding; use org.ngafid.core.util.MD5 for consistency (e.g. DigestOutputStream while writing the temp file, or hash after write with MD5.computeHexHash(Files.newInputStream(dest))).

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,314 @@
...
+                        1,
+                        md5Hash);
+
+                Path archivePath = upload.getArchivePath();
+                Files.createDirectories(archivePath.getParent());
+                Files.move(tempFile, archivePath, StandardCopyOption.REPLACE_EXISTING);
+                tempFile = null;
+
+                // On lock close, Upload auto-publishes to Topic.UPLOAD and kicks off processing.
+                try (Upload.LockedUpload locked = upload.getLockedUpload(connection)) {
```
eightspokes, line 209: Wrap the DB + archive + complete() block in try/catch/finally: on any failure after createNewUpload, delete upload.getArchivePath() if present and remove the upload row  so retries aren’t blocked.

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,314 @@
...
+
+                ctx.status(201).json(new UploadResponse(
+                        upload.getId(),
+                        upload.getStatus().name(),
+                        upload.getFilename(),
+                        upload.getMd5Hash(),
+                        file.size(),
+                        fleetId));
+            }
+        } catch (SQLException e) {
```
eightspokes, line 224: Please use a single error JSON shape for all /api/external/* responses. Right now most errors use ApiError { "error": "..." }, but SQLException returns ErrorResponse { "errorTitle", "errorMessage" }. Use ApiTokenAuth.ApiError for 500s too (generic message to client, full exception in logs only).

ngafid-core/src/main/java/org/ngafid/core/flights/Airframes.java
dcr8024: Just reverted those commits let me know if it looks good

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,342 @@
...
+    public static void bindRoutes(Javalin app) {
+        app.before("/api/external/*", ApiTokenAuth::requireApiToken);
+
+        app.post("/api/external/uploads",            ApiExternalUploadRoutes::postUpload);
+        app.get ("/api/external/uploads",            ApiExternalUploadRoutes::listUploads);
+        app.get ("/api/external/uploads/{uploadId}", ApiExternalUploadRoutes::getUpload);
+    }
+
+    private static void postUpload(Context ctx) {
+    final User user = Objects.requireNonNull(ctx.attribute("user"));
```
travisdesell: format here is incorrect, make sure you run the auto formatter and linter

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
```diff
@@ -0,0 +1,342 @@
...
+            return;
+        }
+        if (file.size() <= 0) {
+            ctx.status(400).json(new ApiTokenAuth.ApiError("Empty file"));
+            return;
+        }
+
+        Path tempFile = null;
+        try {
+            tempFile = Files.createTempFile("ngafid-api-upload-", ".tmp");
```
travisdesell: will this break if two users are uploading via the api at the same time?  the filename should be unique.

travisdesell (changes requested):
Make sure formatting is correct.  Methods all should have appropriate javadoc style comments explaining what they do and what the parameters/return values are.

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: the .createTempFile adds a random set of numbers in between the prefix and suffix 

/tmp/ngafid-api-upload-1234567890123456789.tmp
/tmp/ngafid-api-upload-9876543210987654321.tmp

eightspokes (approved):

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: Changed the logic to follow the sanitize logic

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: Changed Logic as requested

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: Spoke to Travis about this on slack , I will be working on it on a follow up PR as he wants to be able to resume an interrupted upload

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: updated to use the existing logic

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: Updated code to follow that

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
dcr8024: Done

ngafid-www/src/main/java/org/ngafid/www/routes/ApiExternalUploadRoutes.java
eightspokes: Maybe we could use something more descriptive in between prefix and suffix  for traceability/ "debuggability" like upload id or something?

travisdesell: okay great, wasn't sure if it handled that correctly. we could also use the sanitized filename from a few lines above and add that to the filename.

eightspokes: Random number is fine. Disregard my previous comment.

ngafid-db/src/changelogs/01-flights/13-tail-airframe-registry.sql
eightspokes, line 1: This is a workaround. Probably needs to be reverted. The problem is that your branch is behind main. When we merge yours into main , it will work.

PR #338, opened 2026-07-23
Mingfeng/obstacle data class

MingFZ0:
Obstacles are first parsed in the Obstacles class in ngafid-core. Similar to how airports are parsed, it simply reads the obstacle csv file in data-local/obstacles/obstacles_parsed.csv and stores it in a hashmap.

Obstacles are parsed into Events by EventConsumer. When it encounters the obstacle event definitions (-10, -9, -8), it creates the ObstacleEventScanner by passing the flight and the event definition to it.

To make the obstacle events, the event scanner loops through the flight's entries. For each entry, it pulls all of the nearby obstacles within range and creates an event for each to track. 
- If the same tracked obstacle gets pulled up by any of the following flight entries, the event will be updated with the new distance and time. 
- If the event has not been updated in a set period of time, ie., the same obstacle has not been found within range in (hard-coded to be 5 minutes for now), they are inserted into the database with severity as the distance between the obstacle & the aircraft, and horizontal & vertical distance inserted as event metadata.


ngafid-core/src/main/java/org/ngafid/core/obstacles/Obstacles.java
```diff
@@ -0,0 +1,335 @@
...
+
+        try (BufferedReader obstaclesReader = new BufferedReader(new FileReader(Config.OBSTACLES_FILE));) {
+            String line;
+
+            while ((line = obstaclesReader.readLine()) != null) {
+
+                String[] values = line.split(",");
+                int id = Integer.parseInt(values[2]);
+                Double lat = Double.parseDouble(values[10]);
+                Double lon = Double.parseDouble(values[11]);
```
jkarns275:
```
                double lat = Double.parseDouble(values[10]);
                double lon = Double.parseDouble(values[11]);
```

ngafid-core/src/main/java/org/ngafid/core/obstacles/Obstacles.java
```diff
@@ -0,0 +1,335 @@
...
+        
+        if (TEST_MODE) {
+            LOG.info("TEST MODE: skipping reading obstacles files");
+        } else {
+            LOG.info("Obstacle Class was ran");
+
+            try (Connection connection = Database.getConnection();) {
+                
+                // Check if obstacles exist in the database. If not, begin parsing
+                if (VerifyObstaclesInDatabase(connection) == false) {
```
jkarns275: methods should start with a lowercase letter

ngafid-core/src/main/java/org/ngafid/core/obstacles/Obstacles.java
```diff
@@ -0,0 +1,335 @@
...
+        GEO_HASH_TO_OBSTACLES = new HashMap<>();
+        OBJECTID_TO_OBSTACLES = new HashMap<>();
+        OBSTACLE_TYPE_MAP = new HashMap<>();
+        
+        if (TEST_MODE) {
+            LOG.info("TEST MODE: skipping reading obstacles files");
+        } else {
+            LOG.info("Obstacle Class was ran");
+
+            try (Connection connection = Database.getConnection();) {
```
jkarns275, line 44: extra ;

ngafid-data-processor/src/main/java/org/ngafid/processor/events/ObstacleEventScanner.java
```diff
@@ -0,0 +1,202 @@
...
+                                event.getEndTime(), 
+                                event.getStartLine(), 
+                                event.getEndLine(), 
+                                event.getEventDefinitionId(), marked.getTotalDistance()));
+                        tempObstacleDistanceMap.put(obstacleId, marked);
+                    }
+                }
+            }
+
+            ArrayList<Integer> ObstacleIDs = new ArrayList<>();
```
jkarns275: lowercase var name

ngafid-data-processor/src/main/java/org/ngafid/processor/events/ObstacleEventScanner.java
```diff
@@ -0,0 +1,202 @@
...
+            ObstacleIDs.addAll(tempObstacleIDMap.keySet());
+
+            // Loop through all of the tracked obstacle events & check if any of them should be inserted
+            for (int ObstacleID : ObstacleIDs) {
```
jkarns275: lowercase

ngafid-data-processor/src/main/java/org/ngafid/processor/format/FlightBuilder.java
```diff
@@ -140,6 +139,7 @@ protected List<ComputeStep> gatherSteps(Connection connection) {
      * @throws FlightProcessingException if an irrecoverable processing issue is encountered
      */
     public FlightBuilder build(Connection connection) throws FlightProcessingException {
+
```
jkarns275: :angry:

ngafid-data-processor/src/main/kotlin/org/ngafid/processor/EventConsumer.kt
```diff
@@ -216,6 +217,7 @@ class EventConsumer protected constructor(
                 EventScanner(def)
             } else {
                 when (def.id) {
+                    -10, -9, -8 -> ObstacleEventScanner(flight, def)
```
jkarns275, line 220: I see that we will end up running the obstacle event scanner for all three custom event definitions. I wonder if we could just do this in one scan? And classify the event into the proper definition based on the severity.

Also just realizing we should really centralize these definitions / ids in an enum somewhere.

MingFZ0: It would be nice to do it as one scan, but I don't think we have the system for creating an event that has different tiers of severity (high & medium & low). And since each tier have their own conditions, I decided to split them up into 3 separate definitions instead so they show up as different events on the event definition page.

MingFZ0:
Running into a bit of trouble with the obstacle heatmap. Two main points. 
1. Obstacles are inserted into the database within the ObstacleEventScanner because event ids are required after insertion to be recorded in the obstacle_event_key table to link obstacle events and obstacle ids. However, since events are already processed in the scanner, no events are returned to the EventConsumer to avoid duplicate processing. When heatmap points are generated in the EventConsumer, there are no obstacle events for it to process, and therefore no heatmap points are created for obstacles.
> I will try to change obstacles to use the event metadata table instead (sorry Joshua). It is going to be a bit weird, but if there are any future problems, it may be a good idea to change obstacles to rely on a different system rather than Events.
2. The obstacle events are not showing up in the Heatmap as selectable events at all. Even when heatmap points are made, it still does not appear.

MingFZ0:
So currently there are two issues present with obstacles. 

1. Most of the obstacles that are detected are during takeoff and landing when it is near an airport/runway. 
- Originally, I planned to use a static number to ignore obstacles in the ranges of an airport/runway. But after exploring a bit more, it would be a better idea to exclude obstacle detection during the taxiing, takeoff, and landing phases of the flight. The taxiing and takeoff are already implemented in Cesium, just need to find a way to extract that algorithm into obstacles. The landing phase however needs to be figured out separately.

2. 2D map and the 3D Cesium map are not displaying the same obstacle events.
- In the 2D map, the points and the durations of the map make sense as they are clustered around the start and the end of the flight. But in the 3D map, the same obstacle event can stretch out to half of the flight. It seems to be that the optimizations that the 3D map is doing is messing up the obstacle events, which tends to be very short in duration. The implementation of obstacle events seems to be fine, but it doesn't work well with the Cesium map.

